PR #251, opened 2015-03-26
Fixed a couple of integer overflows

ilammy:
I am currently writing an implementation of SRFI 60 for Chibi. During that I was lucky enough with my random test generator to run into some issues with bignums caused by overflows at edge cases. You can see details in the commit messages.

I could not reproduce any more issues with these changes applied when running the SRFI 60 test set taken to extreme (thousands of random 4096-bit numbers) on 32-bit and 64-bit machines available to me. Test generators for tests in the pull request [are located here](https://gist.github.com/ilammy/2a4626c981c91984a0ac).

So it seems that's all for now. But maybe I should consider setting some static code analyzer on Chibi in attempt to find more stuff...

(P.S. Well, I told myself that I want more experience with Chibi's C side, so I got it. Seems fair to me.)


ashinn:
Nice, thanks for catching these corner cases!

Issue #253, opened 2015-04-08
snow-chibi fails to install packages larger than 64K


ilammy:
I ran into this accidentally while trying to install a package with ~~huge~~ comprehensive amount of tests (around 0.5 MB packed). At first I thought that was RSA or tar have slow implementations, but studying the code has shown that signature checks are not run when there is not signature, that gunzip is implemented as a system call, and testing of `(chibi tar)` resulted in sane results for 2 MB tarballs.

[Here is a script that generates dummy snowballs](https://gist.github.com/ilammy/a350386ffe7ec378027a). `example-63.tgz` is successfully installed, but `snow-chibi` decides to take a nap when given `example-64.tgz` or anything larger. I do not see any CPU activity when this happens.

Tapping into `chibi-scheme` process with gdb reliably gives a stack trace like this:

```
Program received signal SIGINT, Interrupt.
0x00007ffff78ac470 in __nanosleep_nocancel ()
    at ../sysdeps/unix/syscall-template.S:82
82	../sysdeps/unix/syscall-template.S: No such file or directory.
#0  0x00007ffff78ac470 in __nanosleep_nocancel ()
    at ../sysdeps/unix/syscall-template.S:82
#1  0x00007ffff78d6b74 in usleep (useconds=<optimized out>)
    at ../sysdeps/unix/sysv/linux/usleep.c:33
#2  0x00007ffff5b52dc5 in sexp_scheduler (ctx=0x7ffff717a060, 
    self=0x7ffff6e3b080, n=1, root_thread=0x7ffff717a060)
    at lib/srfi/18/threads.c:621
#3  0x00007ffff7bc12b5 in sexp_apply1 (ctx=0x7ffff717a060, f=0x7ffff6e3b080, 
    x=0x7ffff717a060) at vm.c:2159
#4  0x00007ffff7bb6961 in sexp_apply (ctx=0x7ffff717a060, proc=0x7ffff7299320, 
    args=0x7ffff707bc60) at vm.c:1009
#5  0x0000000000405ba9 in run_main (argc=4, argv=0x7fffffffe768) at main.c:648
#6  0x0000000000405df0 in main (argc=4, argv=0x7fffffffe768) at main.c:699
```

so it definitely _is_ napping instead of doing something useful.

ilammy:
[...]

PR #254, opened 2015-04-11
Fix for a pipe overflow affecting snow-chibi

ilammy:
This resolves issue #253 (at least [for me](https://travis-ci.org/ilammy/srfi-60/builds/57714289)).

The thing with pipes is that one cannot just write() stuff into it and go away. The trick is to read data fast enough from the read end to be able to write it fast enough into the write end. However, in this case we failed at it and got a deadlock. Chibi has filled gzip's input pipe with compressed data and is blocked in `write-u8` call waiting for gzip to read it. gzip has filled its output pipe with uncompressed data and is blocked in `write()` call waiting for Chibi to read it. A classical deadlock.

Chibi does not seem to provide Scheme procedures for non-blocking IO, so I wrote a utility `feed-pipe` to do pipe transfers properly. It does not block and allows to switch activity when the pipe buffer is full. This resolves the deadlock. If Chibi is going to be blocked on a write into gzip's input pipe, it instead will go and read something from its output pipe, which will unblock gzip, unclog the pipe, and eventually Chibi will be able to write without blocking. The same goes for the read end case.

`feed-pipe` is kinda low-level in that it assumes that the ports are actually _pipe-like_, but it does not check this and fails catastrophically (by deadlocking) if it is not true. So I have wrapped it into more easy-to-use `process-pipe` procedure that runs an arbitrary process using Scheme ports to pipe input and output.

The only bad thing is that `process-pipe` is busy-waiting if the forked process is not cooperating: e.g., leaves its output pipe open without writing anything into it, or does not read what we write into its input pipe. Currently, I do not have ideas on how to fix this. The _proper way_ would be to either use some kind of a select() to peek on pipes, or to spawn two threads for pumping data in and out then just join them.

Also I have fixed an inconsistency in `sexp_write_char()` macro which could return incorrect values sometimes. Well, most of the time its return value is simply ignored, but there are two cases where it isn't: `write-char` ([ASCII](https://github.com/ashinn/chibi-scheme/blob/4dda9230813fb8b26a5825225000f3a718cedd7c/vm.c#L1941-L1942) and [UTF-8](https://github.com/ashinn/chibi-scheme/blob/4dda9230813fb8b26a5825225000f3a718cedd7c/sexp.c#L2287-L2288)) and `write-u8` ([link](https://github.com/ashinn/chibi-scheme/blob/449feb88d3296fb5f41693fd705a2453592b133a/lib/chibi/io/port.c#L356)). And it is expected to have putc() interface there.

P.S. I have read the comment in zlib.scm saying that it should be rewritten in Scheme and avoid piping, but I believe `process-pipe` will be useful anyway.

### DIFF
--- a/include/chibi/sexp.h
+++ b/include/chibi/sexp.h
@@ -1300,7 +1300,7 @@ SEXP_API sexp sexp_push_op(sexp ctx, sexp* loc, sexp x);
 
 #define sexp_read_char(x, p) (sexp_port_buf(p) ? ((sexp_port_offset(p) < sexp_port_size(p)) ? ((unsigned char*)sexp_port_buf(p))[sexp_port_offset(p)++] : sexp_buffered_read_char(x, p)) : getc(sexp_port_stream(p)))
 #define sexp_push_char(x, c, p) ((c!=EOF) && (sexp_port_buf(p) ? (sexp_port_buf(p)[--sexp_port_offset(p)] = ((char)(c))) : ungetc(c, sexp_port_stream(p))))
-#define sexp_write_char(x, c, p) (sexp_port_buf(p) ? ((sexp_port_offset(p) < sexp_port_size(p)) ? ((((sexp_port_buf(p))[sexp_port_offset(p)++]) = (char)(c)), 0) : sexp_buffered_write_char(x, c, p)) : putc(c, sexp_port_stream(p)))
+#define sexp_write_char(x, c, p) (sexp_port_buf(p) ? ((sexp_port_offset(p) < sexp_port_size(p)) ? ((((sexp_port_buf(p))[sexp_port_offset(p)++]) = (char)(c)), (int)(c)) : sexp_buffered_write_char(x, c, p)) : putc(c, sexp_port_stream(p)))
 #define sexp_write_string(x, s, p) (sexp_port_buf(p) ? sexp_buffered_write_string(x, s, p) : fputs(s, sexp_port_stream(p)))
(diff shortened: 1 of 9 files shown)


ashinn:
Another approach is to rely on chibi's threads to handle non-blocking and scheduling with something like:

```scheme
(define (process-pipe-bytevector cmd bvec)
  (call-with-process-io
   cmd
   (lambda (pid proc-in proc-out proc-err)
     (let* ((res #f)
            (th (thread-start!
                 (make-thread
                   (lambda () (set! res (port->bytevector proc-out)))))))
       (write-bytevector bvec proc-in)
       (close-output-port proc-in)
       (waitpid pid 0)
       (thread-join! th #t)
       res))))
```

but feed-pipe may good to have anyway.

ilammy:
Yeah, I don't like busy-waiting too.

Should Chibi support piping (and, by extension, the current implementation of zlib) when it is built without threading support? If yes then `process-pipe` can check for threads with `cond-expand` and use manual IO scheduling via `feed-pipe` only as a fallback.

ilammy:
Also, apparently I was too dumb to notice, but there are `u8-ready?` and `char-ready?` procedures in Scheme, so `feed-pipe` can be written without any implementation-specific C code. For threaded implementation we will need its blocking counterpart (and a name for it).

I will close this pull request for now, remove any unnecessary C code (I don't think we'll lose much performance due to that), integrate Chibi thread usage, and reopen the request with all that stuff done.

Issue #255, opened 2015-04-12
snow-chibi displays poor performance with non-tiny snowballs


ilammy:
[Here is a chart of Snow's running time versus snowball size](https://docs.google.com/spreadsheets/d/1PX4RDSYqBS5aQRPjAa7lMIhMaWeypKL3UJ8sDQ8D5tc/edit?usp=sharing). It is not _that_ bad (could be quadratic), but the factor of this O(_n_) is kinda large. It takes about 2.5 minutes on Travis' VMs to install a 1.5 MB snowball. This is sad.

Running Chibi with `-h 512m` does not help very much, slightly shifing the ratio between the user and system times from 1:3 to 1:4. Anyway, we're talking about 1 MB snowballs here, even the default 32 MB should be enough.

Running Chibi under Valgrind gives the following top calls:

<table><tr><th>Time</th><th>Module</th><th>Function</th></tr>
<tr><td align="right">36.75%</td><td>vm.c</td><td>sexp_apply</td></tr>
<tr><td align="right">10.00%</td><td>gc.c</td><td>sexp_allocated_bytes</td></tr>
<tr><td align="right">6.82%</td><td>gc.c</td><td>sexp_sweep</td></tr>
<tr><td align="right">6.53%</td><td>vm.c</td><td>sexp_apply'2</td></tr>
<tr><td align="right">6.03%</td><td>gc.c</td><td>sexp_try_alloc</td></tr>
<tr><td align="right">5.24%</td><td>gc.c</td><td>sexp_finalize</td></tr>
<tr><td align="right">3.94%</td><td>(libc)</td><td>__memset_x86_64</td></tr>
<tr><td align="right">3.93%</td><td>gc.c</td><td>sexp_reset_weak_references</td></tr>
<tr><td align="right">3.31%</td><td>gc.c</td><td>sexp_alloc</td></tr>
<tr><td align="right">3.26%</td><td>sexp.c</td><td>sexp_make_vector_op</td></tr>
<tr><td align="right">2.96%</td><td>eval.c</td><td>sexp_env_cell_loc</td></tr>
<tr><td align="right">2.53%</td><td>sexp.c</td><td>sexp_alloc_tagged_aux</td></tr>
<tr><td align="right">1.75%</td><td>sexp.c</td><td>sexp_buffered_flush</td></tr>
<tr><td align="right">1.49%</td><td>io.c</td><td>sexp_feed_pipe</td></tr>
[...]

PR #258, opened 2015-04-18
Native SHA-2 implementation (take two)

ilammy:
This is take two on pull request #257 (resolves issue #255). The code was split into the Scheme part which performs input-output in a proper way, and the C part which does actual number crunching.

Native code exports the following interface:

- `(start-sha type)` — allocate a new computation context for the digest of specified type. <br/> The constants for supported digest types are exported as well:
    - `type-sha-224` — SHA-224
    - `type-sha-256` — SHA-256
- `(add-sha-data! sha-context data)` — add a chunk of data into the context. <br/> It supports only bytevectors and strings. I did not include support for single chars and bytes because I did not need it and I doubt anybody ever needing it.
- `(get-sha sha-context)` — retrieve the digest value. <br/> This is a 'destructive' operation in the sense that one cannot add any more data into a context after this call. It was easier to implement it in such way and I did not need partial digests.

I felt like adding a dozen of `make-foo-context`, `add-foo-data!`, `get-foo-digest` is error-prone, and we will have to do digest type checks anyway (e.g., to raise an error instead of returning garbage or causing segfaults when `get-sha-512-digest` is applied to SHA-256 context), so the interface was made easier to use from Scheme and easier to extend to support other SHA-2 digests.

I have also added a bunch of new tests for SHA-2 procedures to check the handling of some important boundary cases. And—lucky me—they exposed a bug in the portable implementation. (Which was fixed, of course.)

Performance figures stay the same in comparison with the portable implementation: around x10000 boost for raw computation without any IO, and x100 boost for the case of input ports.

Oh, and the portable implementation is kept intact in the code for non-Chibi Schemes.


ilammy:
I am not sure on the best workflow for code review using Github's pull requests, but creating yet another 'take three' is surely a hassle, and I am not a fan of keeping 'fixed stuff after review' commits. I have added some fixup commits which are ready for autosquashing. Please don't merge the pull request right away, I will tidy it up in the end, combining these (and possible other) fixups with actual commits.

ilammy:
Okay, I just squashed all fixes into two commits: one moves exact integer definitions to &lt;sexp.h&gt;, and the other one includes all other amendments. I guess the history and comments here would be important.

PR #261, opened 2015-05-02
Add the variable VERSION and remove the uses of 'cat VERSION'

juanfra684:
The Makefile doesn't write correctly the version of chibi-scheme within
the pkgconfig file (spotted by James Turner, OpenBSD developer) because
the VERSION variable doesn't exist.

The uses of 'cat VERSION' are not needed anymore.

/cc @jturner


ashinn:
Thanks for catching this!

PR #265, opened 2015-06-06
Correction to the 8-10th selector aliases in SRFI-1

frerejerome:
Correction to SRFI-1 *selectors.scm* :
The three selector aliases `eighth`, `ninth` and `tenth` were off by one *cdr*.


frerejerome:
**Other commits in patch-1:**
- Added a proper markdown *README* file.
- Added my proposed project logo, cf. [forum post](https://groups.google.com/forum/#!topic/chibi-scheme/SafcUSckmac).

   (Tried to reference the one hosted on synthcode.com, which GitHub superbly ignored.)
- Fixed a compilation warning/error when building without a static library.

ashinn:
Thanks for the fix!

PR #272, opened 2015-06-22
Integrate emscripten build process in Makefile

mnieper:
This is an attempt to integrate the build process for the Emscripten output into the regular Makefile of Chibi Scheme. (So far, external scripts have been provided through https://github.com/mnieper/chibi-scheme.js).

If a recent enough version of Emscripten is installed, `make chibi.js` will first build a static executable under the host system (which is being renamed to `chibi-scheme-emscripten`), which in turn is used to compile object files with the Emscripten tool chain.

Afterwards, navigating to `index.html` shows a little demo page.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Object files
 *.o
+*.bc
 *.ko
 *.obj
 *.elf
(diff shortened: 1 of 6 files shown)


mnieper:
Moved everything js-related into the separate directory `js/`.

ashinn:
LGTM, thanks!

Issue #280, opened 2015-10-04
libchibi-scheme.so.0 not found when using a prefix build/install on Linux


mcalan:
When building/installing with PREFIX=/opt/chibi I get the following linker errors:

    $ chibi-scheme
    chibi-scheme: error while loading shared libraries: libchibi-scheme.so.0: cannot open shared object file: No such file or directory
    $ ldd /opt/chibi/bin/chibi-scheme 
	     linux-vdso.so.1 (0x00007ffd5502f000)
	     libchibi-scheme.so.0 => not found
	     libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f71b856b000)
	     /lib64/ld-linux-x86-64.so.2 (0x00007f71b8914000)
    $ ldd /opt/chibi/lib/chibi/chibi/net.so 
	     linux-vdso.so.1 (0x00007ffec7335000)
	     libchibi-scheme.so.0 => not found
	     libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007f792c330000)
	     /lib64/ld-linux-x86-64.so.2 (0x00007f792c8df000)

mcalan:
Proposed fix in pull request #281.

ashinn:
Thanks for the report and patch!  However, this has been brought up before.  I've always used LD_LIBRARY_PATH when working with non-standard install paths.  It offers more control, avoiding hard-coding paths which could break when files are moved, or from chroot, etc.

Moreover Debian advises against using rpath: https://wiki.debian.org/RpathIssue

mcalan:
Fair enough, as you say it's easy enough to work around with an alias. Thanks for the link it was a good read.

cipriancraciun:
I've started a thread on the Chibi mailing list about this issue, with the subject `Regarding '-rpath' during build, or how it enables usability (and in relation with issue #280 on GitHub)`

  https://groups.google.com/d/topic/chibi-scheme/B8KpwuVOH48/discussion

PR #281, opened 2015-10-05
Fix dynamic linking on Linux when installed outside of the current linker path

mcalan:
Adds rpath when dynamically linking the interpreter and shared libraries.


mcalan:
Proposed fix for issue #280.

mcalan:
Closing per discussion in #280.

PR #289, opened 2015-11-03
Remove unnecessary variable referencing

frerejerome:
Those variables are now correctly detected when defined in the parent makefile.

---
Hello Alex,

I'm working on a simple makefile to cleanly build Chibi-Scheme with the MinGW-W64 toolchain (distributed by Stephan T. Lavavej at http://nuwen.net/mingw.html) on Windows 10.

The shared library and the interpreter are done (with a few warnings on type boundaries), and I'm now looking into the module system. Windows isn't keen on providing the required threads, sockets and system calls. I'll see what is possible to port over with the Windows standard libraries, and maybe with a cross-platform solution like libuv (http://github.com/libuv/libuv).

- Frère Jérôme

### DIFF
--- a/Makefile.detect
+++ b/Makefile.detect
@@ -110,7 +110,7 @@ endif
 ########################################################################
 # Check for headers (who needs autoconf?)
 
-ifndef $(SEXP_USE_NTP_GETTIME)
+ifndef SEXP_USE_NTP_GETTIME
 SEXP_USE_NTP_GETTIME := $(shell echo "main(){struct ntptimeval n; ntp_gettime(&n);}" | gcc -fsyntax-only -include sys/timex.h -xc - >/dev/null 2>/dev/null && echo 1 || echo 0)
 endif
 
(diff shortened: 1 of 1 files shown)


frerejerome:
Commenting on the compilation warnings, they are all related to the types `sexp_sint_t` and `sexp_uint_t`. They exceed on Windows the boundaries of a `long`.

One warning would need a simple cast to `(long)` in _sexp.c_ (on line 2876), or a `%lld` format, but three others are less trivial. They are located in _bignum.c_ on lines 528, 529 & 541 and involve a shift overflow on `(1uL<<(sizeof(sexp_uint_t)*4)`.

Would it be safe for the rest of the bignum logic to apply this shift to a bigger integer `1uLL`?

ashinn:
Thanks!  The formats I'm not worried about, but we can remove warnings if you want.
Best is with macros like PRIu64.

The bignum warning I'll have to take a look at.

ashinn:
First let's be sure our types are correct: sexp_[su]int_t should be the size of a pointer, i.e. 64 bits on a 64-bit platform.

Is SEXP_64_BIT correctly detected?

Is sexp_[su]int_t bound to a 64-bit type?  chibi-scheme is making the assumption that long on a 64-bit system is 64 bits, but this seems to be unix specific - on windows long is always 32 bits, and we need an appropriate ifdef in sexp.h to use long long for 64 bits.  We mustn't use long long on unix, as that would become 128 bits.

We also need appropriate definitions for sexp_l[su]int_t in bignum.h, assuming [su]int128_t are not available.

Then, if and only if the compiler doesn't promote the literal types correctly, we would need to change the code from 1uL to ((sexp_uint_t)1) to get the correct size.

frerejerome:
Thank you for checking this through. I asserted sexp_[su]int_t on compilation and they are both correctly defined as 64-bit integers when SEXP_64_BIT is used.

These warnings ultimately come from the `1uL` value which on Windows only amount to 4 bytes.
Replacing its occurrences in bignum.c with a more sensible `(sexp_uint_t)1` fixed all warnings.

PR #291, opened 2015-11-19
Patch-2 for MinGW

frerejerome:
Allows dynamic loading of libraries on Windows, provided that `sexp sexp_init_library()` is properly exported. Could `SEXP_API` be used in this case?
If so, it would be necessary to go through the tree and add this prefix to each library *C* file.

### DIFF
--- a/include/chibi/sexp.h
+++ b/include/chibi/sexp.h
@@ -690,7 +690,7 @@ SEXP_API sexp sexp_make_flonum(sexp ctx, float f);
 #define sexp_flonump(x)      (sexp_check_tag(x, SEXP_FLONUM))
 #define sexp_flonum_value(f) ((f)->value.flonum)
 #define sexp_flonum_bits(f) ((f)->value.flonum_bits)
-sexp sexp_make_flonum(sexp ctx, double f);
+SEXP_API sexp sexp_make_flonum(sexp ctx, double f);
 #endif
 
 #define sexp_typep(x)       (sexp_check_tag(x, SEXP_TYPE))
--- a/include/chibi/sexp.h
+++ b/include/chibi/sexp.h
@@ -21,6 +21,13 @@ extern "C" {
 #define sexp_isdigit(x) ((isdigit)((int)(x)))
 #define sexp_tolower(x) ((tolower)((int)(x)))
 #define sexp_toupper(x) ((toupper)((int)(x)))
+#if SEXP_USE_DL
+#define RTLD_DEFAULT 0
+#define RTLD_LAZY    0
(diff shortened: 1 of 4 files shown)


lib/chibi/ast.c
```diff
@@ -429,11 +434,11 @@ static sexp sexp_set_atomic (sexp ctx, sexp self, sexp_sint_t n, sexp new_val) {
 #endif
 
 sexp sexp_thread_list (sexp ctx, sexp self, sexp_sint_t n) {
-  sexp ls;
   sexp_gc_var1(res);
   sexp_gc_preserve1(ctx, res);
   res = SEXP_NULL;
 #if SEXP_USE_GREEN_THREADS
+  sexp ls;
```
ashinn: This is not portable C - I want the code to compile on various pre-C99 compilers including plan9.

lib/chibi/ast.c
```diff
@@ -530,7 +535,7 @@ static sexp sexp_abort (sexp ctx, sexp self, sexp_sint_t n, sexp value) {
 #define sexp_define_type(ctx, name, tag) \
   sexp_env_define(ctx, env, sexp_intern(ctx, name, -1), sexp_type_by_index(ctx, tag));
 
-sexp sexp_init_library (sexp ctx, sexp self, sexp_sint_t n, sexp env, const char* version, const sexp_abi_identifier_t abi) {
+SEXP_API sexp sexp_init_library (sexp ctx, sexp self, sexp_sint_t n, sexp env, const char* version, const sexp_abi_identifier_t abi) {
```
ashinn: SEXP_API is on most platforms "extern" which doesn't make sense on a definition.

lib/chibi/ast.c
frerejerome: Understood.

include/chibi/sexp.h
```diff
@@ -683,7 +690,7 @@ SEXP_API sexp sexp_make_flonum(sexp ctx, float f);
 #define sexp_flonump(x)      (sexp_check_tag(x, SEXP_FLONUM))
 #define sexp_flonum_value(f) ((f)->value.flonum)
 #define sexp_flonum_bits(f) ((f)->value.flonum_bits)
-sexp sexp_make_flonum(sexp ctx, double f);
```
frerejerome: While `SEXP_API` might not apply in other instances, wouldn't it be needed here? If not, I'll simply withdraw the pull request and look for another way to let the linker know about those functions (Windows DLLs need these `extern` declarations).

ashinn: Yes, here and any other declarations are fine.

PR #297, opened 2015-12-22
Fixed all issues reported by XCode static code anayzer, run with 'Pro…

cwds:
…duct/Analyze'.  Fixes need to be reviewed and/or corrected by original code author if possible

### DIFF
--- a/bignum.c
+++ b/bignum.c
@@ -237,6 +237,10 @@ sexp sexp_bignum_fxrem (sexp ctx, sexp a, sexp_sint_t b) {
   b0 = (b >= 0) ? b : -b;
   for (i=len-1; i>=0; i--) {
     n = (n << sizeof(sexp_uint_t)*8) + data[i];
+    // ISSUE_296 - Division by zero, assigned line 237 if b >= 0
+    if (b0 == 0) {
+      return sexp_read_error(ctx, "division by zero", SEXP_NULL, NULL);
+    }
     q = n / b0;
     n -= (sexp_luint_t)q * b0;
   }
@@ -303,6 +307,10 @@ sexp sexp_write_bignum (sexp ctx, sexp a, sexp out, sexp_uint_t base) {
   sexp_gc_preserve2(ctx, b, str);
   b = sexp_copy_bignum(ctx, NULL, a, 0);
   sexp_bignum_sign(b) = 1;
+  // ISSUE_296 - Division by zero, assigned line 237 if b >= 0
+  if (lg_base == 0) {
+    return sexp_read_error(ctx, "division by zero", SEXP_NULL, NULL);
+  }
   i = str_len = (sexp_bignum_length(b)*sizeof(sexp_uint_t)*8 + lg_base - 1)
     / lg_base + 1;
   str = sexp_make_string(ctx, sexp_make_fixnum(str_len),
--- a/eval.c
+++ b/eval.c
@@ -236,7 +236,8 @@ sexp sexp_extend_synclo_env (sexp ctx, sexp env) {
       sexp_env_renames(e2) = sexp_env_renames(e1);
 #endif
     }
-    sexp_env_parent(e2) = sexp_context_env(ctx);
+    // ISSUE_296 - Dereference of null pointer, stored line 231
+    if (e2) { sexp_env_parent(e2) = sexp_context_env(ctx); }
   }
   sexp_gc_release1(ctx);
   return e;
@@ -259,9 +260,10 @@ static sexp sexp_flatten_dot (sexp ctx, sexp ls) {
 
 int sexp_param_index (sexp ctx, sexp lambda, sexp name) {
   sexp ls;
-  int i;
+  int i = 0;
   while (1) {
-    i = 0;
+    // ISSUE_296 - Value stored to 'i' is never used
+    //i = 0;
     ls = sexp_lambda_params(lambda);
     for (i=0; sexp_pairp(ls); ls=sexp_cdr(ls), i++)
       if (sexp_car(ls) == name)
@@ -1852,7 +1854,8 @@ sexp sexp_type_slot_offset_op (sexp ctx , sexp self, sexp_sint_t n, sexp type, s
     v = &sexp_type_slots(type);
     len = 1;
   }
-  len = sexp_vectorp(cpl) ? sexp_vector_length(cpl) : 1;
+  // ISSUE_296 - len never used (set twice with same thing)
+  //len = sexp_vectorp(cpl) ? sexp_vector_length(cpl) : 1;
   for (i=0; i<len; i++)
     for (slots=sexp_type_slots(v[i]); sexp_pairp(slots); slots=sexp_cdr(slots), offset++)
       if (sexp_car(slots) == slot)
--- a/gc.c
+++ b/gc.c
@@ -362,7 +362,8 @@ sexp sexp_finalize (sexp ctx) {
       for (r=q->next; r && ((char*)r<(char*)p); q=r, r=r->next)
         ;
       if ((char*)r == (char*)p) { /* this is a free block, skip it */
-        p = (sexp) (((char*)p) + r->size);
+        // ISSUE_296 - Access to 'size' dereference NULL 'r'
+        p = (sexp) (((char*)p) + (r ? r->size : 0));
         continue;
       }
       size = sexp_heap_align(sexp_allocated_bytes(ctx, p));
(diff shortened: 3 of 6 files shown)


bignum.c
```diff
@@ -237,6 +237,10 @@ sexp sexp_bignum_fxrem (sexp ctx, sexp a, sexp_sint_t b) {
   b0 = (b >= 0) ? b : -b;
   for (i=len-1; i>=0; i--) {
     n = (n << sizeof(sexp_uint_t)*8) + data[i];
+    // ISSUE_296 - Division by zero, assigned line 237 if b >= 0
+    if (b0 == 0) {
+      return sexp_read_error(ctx, "division by zero", SEXP_NULL, NULL);
```
ashinn: This should be checked outside the loop, and return a user exception, not read error.  Actually, it should be checked in sexp_remainder along with the divide by SEXP_ONE case.

eval.c
```diff
@@ -236,7 +236,8 @@ sexp sexp_extend_synclo_env (sexp ctx, sexp env) {
       sexp_env_renames(e2) = sexp_env_renames(e1);
 #endif
     }
-    sexp_env_parent(e2) = sexp_context_env(ctx);
+    // ISSUE_296 - Dereference of null pointer, stored line 231
+    if (e2) { sexp_env_parent(e2) = sexp_context_env(ctx); }
```
ashinn: Can only be null if the alloc above failed, in which case we should return OOM.
It's been a standing TODO to go through and catch all possible OOM cases.

eval.c
```diff
@@ -259,9 +260,10 @@ static sexp sexp_flatten_dot (sexp ctx, sexp ls) {
 
 int sexp_param_index (sexp ctx, sexp lambda, sexp name) {
   sexp ls;
-  int i;
+  int i = 0;
   while (1) {
-    i = 0;
+    // ISSUE_296 - Value stored to 'i' is never used
+    //i = 0;
```
ashinn: That comment is confusing... 'i' is indeed used, it's the return value.
Moreover moving the initialization outside the loop breaks the logic.

eval.c
```diff
@@ -1852,7 +1854,8 @@ sexp sexp_type_slot_offset_op (sexp ctx , sexp self, sexp_sint_t n, sexp type, s
     v = &sexp_type_slots(type);
     len = 1;
   }
-  len = sexp_vectorp(cpl) ? sexp_vector_length(cpl) : 1;
+  // ISSUE_296 - len never used (set twice with same thing)
+  //len = sexp_vectorp(cpl) ? sexp_vector_length(cpl) : 1;
```
ashinn: Yep, that's a duplicated computation.

gc.c
```diff
@@ -362,7 +362,8 @@ sexp sexp_finalize (sexp ctx) {
       for (r=q->next; r && ((char*)r<(char*)p); q=r, r=r->next)
         ;
       if ((char*)r == (char*)p) { /* this is a free block, skip it */
-        p = (sexp) (((char*)p) + r->size);
+        // ISSUE_296 - Access to 'size' dereference NULL 'r'
+        p = (sexp) (((char*)p) + (r ? r->size : 0));
```
ashinn: not possible

gc.c
```diff
@@ -404,7 +405,8 @@ sexp sexp_sweep (sexp ctx, size_t *sum_freed_ptr) {
       for (r=q->next; r && ((char*)r<(char*)p); q=r, r=r->next)
         ;
       if ((char*)r == (char*)p) { /* this is a free block, skip it */
-        p = (sexp) (((char*)p) + r->size);
+        // ISSUE_296 - Access to 'size' dereference null 'r'
+        p = (sexp) (((char*)p) + (r ? r->size : 0));
```
ashinn: not possible

gc.c
```diff
@@ -712,7 +714,8 @@ void sexp_offset_heap_pointers (sexp_heap heap, sexp_heap from_heap, sexp* types
       for ( ; q && ((char*)q < (char*)p); q=q->next)
         ;
       if ((char*)q == (char*)p) { /* this is a free block, skip it */
-        p = (sexp) (((char*)p) + q->size);
+        // ISSUE_296 - Access to field 'size' dereference NULL 'q'
+        p = (sexp) (((char*)p) + (q ? q->size : 0));
```
ashinn: not possible

bignum.c
cwds: The divide by zero case seems to be caught in sexp_bignum_quot_rem() function, so on inspection it appears 'b' could never be zero in this function.  I didn't see any obvious test case which would invoke this condition.  A cut and paste from sexp_bignum_quot_rem() silences the warning, and seems to be the way to throw a user exception in this case:

```c
  b0 = (b >= 0) ? b : -b;
  if (b0 == 0) {
    return sexp_xtype_exception(ctx, NULL, "divide by zero", a);
  }
```

eval.c
cwds: Looking through the Static Analyzer report, it seems to be complaining about the case where env may be NULL (probably impossible) in the function call.  In that case the for loop never iterates, e1 and e2 are both NULL, and the sexp_env_parent(e2) is dereferencing a NULL pointer.

Put in what I believe is an OOM check to silence the warning, which probably isn't an accurate error for this case:

```c
    if (!e2) { return sexp_global(ctx, SEXP_G_OOM_ERROR); }
    sexp_env_parent(e2) = sexp_context_env(ctx);
```

As far as checking all possible OOM cases, is there a reason why this couldn't be done in the allocator (`sexp_alloc_type`) and have it return the SEXP exception rather than at the result sites?

eval.c
cwds: This is a pretty useless static analyzer warning.  The value of i is initialized upon the first pass of the for loop, which overrides the initialization value the analyzer is complaining about a few lines before.  It's just redundant code.  Initializing outside the loop also performs no function (it would be redundant) as the first thing the for loop does is set i to 0; but initializing things in the declaration is often considered good practice should never break anything; otherwise the logic would be depending on the initialized value.

Removed the redundant i = 0 statement to get rid of the warning.
```c
 while (1) {
    ls = sexp_lambda_params(lambda);
    for (i=0; sexp_pairp(ls); ls=sexp_cdr(ls), i++)
      if (sexp_car(ls) == name)
        return i;
...
```

eval.c
cwds: Removed redundant code.

gc.c
cwds: Whatever the static analyzer was complaining about it had 18 arrows to describe the situation.  Changed the code as follows to silence the warning - should have no effect as case is not possible.

```c
        p = (sexp) (((char*)p) + (r ? r->size : 0));
```

gc.c
cwds: Left change in to silence warning; isn't hurting anything if case not possible.

gc.c
cwds: Left change in to silence warning; isn't hurting anything if case not possible.

cwds:
Thanks for commenting and clarifying the suggested changes to remove warnings.  All corrections were submitted as individual checkin's on my chibi-scheme fork at 
```
https://github.com/cwds/chibi-scheme
```

Hopefully the checkin comments are clear should you choose to incorporate these mostly trivial fixes.  The Clang Static Analyzer falls silent after these changes are applied; none of them seem particularly significant.

In addition, the aformentioned fork also adds the build option SEXP_USE_STATIC_SLIBS to Chibi.  When enabled, it allows one to compile in all of the .scm and .sld files needed from the library into the data segment of your executable, allowing for an end product with no dependancies.  The file FORK.md explains the rational and implementation.  The feature may prove useful to a wider audience, and should in no way affect the existing code when disabled.  There are certainly better ways to do this (like at least storing the compiled byte code instead of the raw text, or even compressing the text) but the implementation is simple and adequate for it's intended use.

Once again, thanks for both providing and maintaining this codebase.  Depending on your plans for the codebase, I may be able to help realize those goals while using the code for a current project.  In any event, the file FORK.md explains the changes planned for the ongoing project.

gc.c
ashinn: r can easily be NULL if it's the last free block.  However, we're inside r == p, and p can never be NULL.  It's easy to see how the static analyzer might think this possible, though, but you must understand such tools are fallible.  If p ever were NULL here it would mean we have heap corruption, at which point all is lost and _nothing_ we do here matters.

You claim adding this check isn't "hurting" anything, but it is.  For one it's just that much extra code to read and understand.  Moreover, for the close reader it can add a huge amount of confusion - such a check suggests the heap may be partially corrupt yet we're expected to somehow recover from it here.  Except that the change in question doesn't recover, but just sets p to itself and enters an infinite loop.

Finally, this is in the inner loop of an inner loop that is scanning the entire heap - it is highly performance sensitive.  While there may be little difference on x86 with its branch prediction, it can be a significant slowdown on ARM.

This extra check helps nothing and hurts a lot.

eval.c
ashinn: Yes, this looks fine.

ashinn:
Thanks!  I'll take a look at you're SEXP_USE_STATIC_SLIBS change, which has been on my todo list for a while.

PR #322, opened 2016-04-08
Exclude unsupported functions in Emscripten version

mnieper:
The ‘(chibi system)’ library uses some C functions that are currently not available in Emscripten compiled code so they have to be omitted when the target is ‘emscripten’.

### DIFF
--- a/lib/chibi/system.sld
+++ b/lib/chibi/system.sld
@@ -1,9 +1,9 @@
 
 (define-library (chibi system)
   (export get-host-name
-          user-information user? user-name user-password
+          user? user-name user-password
           user-id user-group-id user-gecos user-home user-shell
-          group-information group-name group-password group-id
+          group-name group-password group-id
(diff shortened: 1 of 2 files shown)


lib/chibi/system.sld
```diff
@@ -12,12 +12,16 @@
...
-              (getpwnam_r user (make-string 1024))
-              (getpwuid_r user (make-string 1024)))))
-   (define (group-information group)
-     (car (if (string? group)
-              (getgrnam_r group (make-string 1024))
-              (getgrgid_r group (make-string 1024)))))))
+  (cond-expand
+   (emscripten)
+   (else
+    (export user-information group-information)
```
ashinn: Don't you need to remove these from the above export?

mnieper: Removed them. Didn't catch the warning.

PR #341, opened 2016-05-17
fix struct for function args

### LINKED ISSUE #330: chibi-ffi struct in function args no effect
test1.stub
```
(define-c-struct point
(int x point-x)
(int y point-y))

(define-c void test (point))
```
test2.stub
```
(define-c-struct point
(int x point-x)
(int y point-y))

(define-c void test ((struct point)))
```
`chibi-ffi test1.stub test1.c`
`chibi-ffi test2.stub test2.c`

`diff test1.c test2.c`

Kipples:
resolves #330

### DIFF
--- a/tools/chibi-ffi
+++ b/tools/chibi-ffi
@@ -764,7 +764,7 @@
             (void*? (void-pointer-type? type)))
         (cond
          ((or ctype void*?)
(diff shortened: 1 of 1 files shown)


ashinn:
Thanks!  Merged along with a unit test.

PR #342, opened 2016-05-18
Modify makefiles to make LDCONFIG optional, support systems with no s…

vmanis:
…ymlinks.

  - Makefile.libs: changed definition of LN to LN -sf, so can be overridden
    with LN=cp on systems with no symlinks; introduced LDCONFIG, so can be
    overridden if desired.

  - Makefile: changed uses of $(LN) -sf to $(LN); replaced two occurrences
    of ldconfig by $(LDCONFIG); suppress install of $(IMAGE_FILES) if variable
    is empty.

Note: the IMAGE_FILES change was to enable Chibi to be compiled on GNURoot+Android,
and can be reasonably reverted if an alternate way of dealing with image files
is chosen.


vmanis:
As per discussion on Chibi-Scheme mailing list.

PR #354, opened 2016-06-05
Handle imaginary values of inverse of sine and cosine on the real axis.

### LINKED ISSUE #167: inverse sine and cosine should promote to complex instead of +nan.0
```
What steps will reproduce the problem?
> (asin -4) => +nan.0

What is the expected output? What do you see instead?
> (asin -4)
-1.5707963267948968+2.0634370688955608i

For inputs outside the range [-pi..pi] we need to
resort to complex values.
```

Original issue reported on code.google.com by `alexsh...@gmail.com` on 25 Dec 2012 at 10:29

mnieper:
Fixes #167.

### DIFF
--- a/eval.c
+++ b/eval.c
@@ -1487,12 +1487,35 @@ sexp sexp_register_optimization (sexp ctx, sexp self, sexp_sint_t n, sexp f, sex
     return sexp_make_flonum(ctx, cname(d));                             \
   }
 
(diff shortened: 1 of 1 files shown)


mnieper:
Note that #167 incorrectly speaks of the critical points -pi and pi. In fact, the branch cuts of the inverse sine and cosine start at -1 and 1.

PR #414, opened 2017-06-04
Remove argv[0] from the arguments of main_symbol.

### LINKED ISSUE #413: SRFI-22: the arguments of main_symbol contains argv[0] incorrectly
From [SRFI-22](https://srfi.schemers.org/srfi-22/srfi-22.html):
> Upon invocation of a script, the Scheme system calls a procedure named main with one argument, a list of strings containing the Unix command-line arguments to the script, i.e. the elements of the argv vector of the Scheme script interpreter process from index 1 on. Thus, the first element of the list is the name of the script.

However, Chibi passes `argv[0]`. Thus,
```shell
$ cat test.scm
#! /usr/bin/env scheme-r7rs
(import (scheme r5rs))
(define (main args)
  (display args)
  (newline))
$ chibi-scheme -r test.scm
("chibi-scheme" "test.scm")
```
Another example: the `cat` implementation in the "Example" section of the SRFI:
```scheme
#! /usr/bin/env scheme-r7rs
(import (scheme r5rs))

(define (main arguments)
  (for-each display-file (cdr arguments))
  0)

(define (display-file filename)
  (call-with-input-file filename
    (lambda (port)
      (let loop ()
        (let ((thing (read-char port)))
          (if (not (eof-object? thing))
            (begin
              (write-char thing)
              (loop))))))))
```
When `chibi-scheme -r ./cat ./cat` is executed, the source code is displayed twice. It is not the expected behavior IMHO.

I confirmed with b52711cac8fca57b9f10ea8c6746069cca608dd1 .

omasanori:
While the `(command-line)` in `(scheme process-context)` contains the command name (`argv[0]`), SRFI 22 specifies that the interpreter passes the command-line arguments except `argv[0]` to the script.

Fix #413.

### DIFF
--- a/lib/chibi/net/http-server.scm
+++ b/lib/chibi/net/http-server.scm
@@ -522,4 +522,4 @@
       (verbose? boolean (#\v "verbose"))))
     ,run-app))
 
-(define (main args) (run-application app-spec args))
+(define (main args) (run-application app-spec))
--- a/lib/chibi/tar.scm
+++ b/lib/chibi/tar.scm
@@ -280,18 +280,17 @@
         res))))
 
 (define (main args)
-  (let ((args (cdr args)))
-    (cond
-     ((equal? "t" (car args))
-      (for-each (lambda (f) (write-string f) (newline)) (tar-files (cadr args))))
-     ((equal? "x" (car args))
-      (if (tar-safe? (cadr args))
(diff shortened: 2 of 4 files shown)


omasanori:
Updated.

ashinn:
Thanks!

PR #430, opened 2017-08-01
Implement SRFI 139

### LINKED ISSUE #383: Implement SRFI 139: Syntax Parameters
Rationale: One cannot write a portable library to implement syntax parameters so support by the core system is needed.

mnieper:
Simple implementation of SRFI 139: Syntax parameters (fixes #383).

- This implementation makes no difference between `define-syntax` and `define-syntax-parameter`, so any keyword may be syntax-parameterized.
- `syntax-parameterize` first adjust the cells containing the syntax parameters, and then emits the body.
- At the end of the body, another macro expansion resets the cells again.
- Should be safe under the assumption that using non-local `call/cc` in macro transformers is not well-defined behavior, and that macro expansion is not interrupted by a thread switch.


ashinn:
Thanks!  Looks awesome, will review soon before merging.

PR #432, opened 2017-08-06
More explicit ways to add compiler flags in chibi-ffi .stub files

briancaine:
My first commit is probably pretty uncontroversial. Just a global you can add additional compiler flags to.

The second one I just added a function to add flags from a command line script like `pkg-config` or (in my case) `sdl2-config`.

### DIFF
--- a/tools/chibi-ffi
+++ b/tools/chibi-ffi
@@ -41,6 +41,7 @@
 (define *consts* '())
 (define *inits* '())
 (define *clibs* '())
+(define *cflags* '())
 (define *tags* '())
 (define *open-namespaces* '())
 (define *c++?* #f)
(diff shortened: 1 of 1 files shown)


tools/chibi-ffi
```diff
@@ -414,6 +415,23 @@
 (define (c-link lib)
   (set! *clibs* (cons lib *clibs*)))
 
+(define (c-flags-from-script cmd)
+  (eval '(import (chibi process)
+                 ; lest I clobber the existing string-split:
+                 (prefix (chibi string) imported:)
+                 (chibi char-set full)) ; for char-set:whitespace
```
ashinn: I think it's simpler to extend the existing string-split to support predicates and use char-whitespace?.

Issue #416, opened 2017-06-12
Invoke `main` procedure by default when argv[0] is `scheme-r7rs` defined by SRFI 22


omasanori:
Now Chibi invokes the `main` procedure when `-r` option is passed without an argument. For scripting purpose, however, it is useful if there is a way to invoke `main` without passing any option for using with shebang.

`#!/usr/bin/env chibi-scheme -r` is not suitable for this purpose since multiple arguments in shebang is not portable. In fact, the shebang line above is interpreted as passing the script to "chibi-scheme -r" command on Linux and it does not work as expected.

My proposal is invoking `main` by default when the interpreter is executed as `scheme-r7rs`, the name defined in SRFI 22, instead of changing the default behavior which introduces incompatibility with current behavior (and R7RS semantics; Shiro Kawai told me this point when I considered the idea.)

ashinn:
Do you mean scheme-script?  Sounds reasonable to me, patches welcome otherwise I may not get around to it soon.

omasanori:
> Do you mean scheme-script?

No, I mean `scheme-r7rs` as specified in the third paragraph, "Semantics" section of SRFI 22:

> scheme-rnrs expects code written in RnRS Scheme. scheme-ieee-n-y expects code written in IEEE n-y Scheme. Specifically, scheme-r4rs expects code written in R4RS Scheme, scheme-r5rs expects code written in R5RS Scheme, and scheme-ieee-1178-1990 expects code written in IEEE 1178-1990 Scheme. scheme-srfi-0 expects code written in R5RS Scheme using the extensions specified in SRFI 0. scheme-srfi-7 expects code written in R5RS Scheme using the extensions specified in SRFI 7. 

> Sounds reasonable to me, patches welcome otherwise I may not get around to it soon.

Thanks! I will try.

PR #435, opened 2017-08-11
[RFC] Treat `scheme-r7rs` command name as in SRFI 22.

omasanori:
I wrote a patch for #416 but I have no idea how we test the feature without making some duplicated code in `tests/run`. Would you give me some advice, or is it acceptable to make a clone of `command-line-tests.sh
`?

### DIFF
--- a/main.c
+++ b/main.c
@@ -304,6 +304,17 @@ sexp run_main (int argc, char **argv) {
   args = SEXP_NULL;
   env = NULL;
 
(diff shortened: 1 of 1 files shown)


ashinn:
I think it's fine to add extra tests to command-line-tests.sh.  You'll want to add a symbolic link to chibi-scheme called scheme-r7rs.

PR #436, opened 2017-08-13
Added support for fixed-length arrays in structs to chibi-ffi

charles-l:
For example:
```
   ;; struct v2 {
   ;;  float vals[2];
   ;; };

   (define-c-struct v2
     predicate: v2?
     constructor: (make-v2 vals)
     ((array float 2) vals v2-vals v2-vals-set!)))
```

I'm using a vector for the array. Let me know if you want it to be a list instead.


charles-l:
Oops - realized I broke fixed length arrays for non-struct static structs. Closing until I fix.

PR #437, opened 2017-11-05
chibi-genstatic: Remove inline? override

okuoku:
Remove `inline?` binding here; it was effectively no-op'ed `--inline` option.

I guess this override is unintentional, or, don't we need `--inline` any longer perhaps?


tools/chibi-genstatic
```diff
@@ -425,8 +425,7 @@
           (set! *features* features))
       (let* ((c-libs+c-files (find-c-libs includes excludes cfiles))
              (c-libs (car c-libs+c-files))
-             (c-files (cdr c-libs+c-files))
-             (inline? #t))
```
ashinn: I can't remember why I did this, but we shouldn't change the default.  We should make inline? default to true above and provide a --no-inline option or similar.

PR #438, opened 2017-11-05
Win32: Fix win32 port

okuoku:
Try to fix win32 port. Now it runs both on Win32/Win64.

Win64 port depends on 128bits arithmetic thus it does not run on MSVC.

- `sexp_make_integer` and `sexp_make_unsigned_integer`: I guess lsint/luint types are not intended to be used outside of bignum implementation. Use sint/uint types to be compatible with non-bignum portions.
- Various LLP64 support.

### DIFF
--- a/bignum.c
+++ b/bignum.c
@@ -531,8 +531,8 @@ sexp sexp_bignum_quot_rem (sexp ctx, sexp *rem, sexp a, sexp b) {
            << (sizeof(sexp_uint_t)*8))
           + sexp_bignum_data(b1)[blen-2]);
     if (alen > 2 && blen > 2 &&
-        sexp_bignum_data(a1)[alen-1] < (1uL<<(sizeof(sexp_uint_t)*4)) &&
-        sexp_bignum_data(b1)[blen-1] < (1uL<<(sizeof(sexp_uint_t)*4))) {
+        sexp_bignum_data(a1)[alen-1] < ((sexp_luint_t)1<<(sizeof(sexp_uint_t)*4)) &&
+        sexp_bignum_data(b1)[blen-1] < ((sexp_luint_t)1<<(sizeof(sexp_uint_t)*4))) {
       dn = (dn << (sizeof(sexp_uint_t)*4))
         + (sexp_bignum_data(a1)[alen-3] >> (sizeof(sexp_uint_t)*4));
       dd = (dd << (sizeof(sexp_uint_t)*4))
@@ -544,8 +544,8 @@ sexp sexp_bignum_quot_rem (sexp ctx, sexp *rem, sexp a, sexp b) {
              << (sizeof(sexp_uint_t)*8))
             + sexp_bignum_data(a1)[alen-2]);
       dd = sexp_bignum_data(b1)[blen-1];
-      if (sexp_bignum_data(a1)[alen-1] < (1uL<<(sizeof(sexp_uint_t)*4)) &&
-          sexp_bignum_data(b1)[blen-1] < (1uL<<(sizeof(sexp_uint_t)*4))) {
+      if (sexp_bignum_data(a1)[alen-1] < ((sexp_luint_t)1<<(sizeof(sexp_uint_t)*4)) &&
+          sexp_bignum_data(b1)[blen-1] < ((sexp_luint_t)1<<(sizeof(sexp_uint_t)*4))) {
         dn = (dn << (sizeof(sexp_uint_t)*4))
           + (sexp_bignum_data(a1)[alen-3] >> (sizeof(sexp_uint_t)*4));
         dd = (dd << (sizeof(sexp_uint_t)*4))
--- a/include/chibi/sexp.h
+++ b/include/chibi/sexp.h
@@ -17,8 +17,9 @@ extern "C" {
 #include "chibi/features.h"
 #include "chibi/install.h"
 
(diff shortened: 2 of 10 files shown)


bignum.c
```diff
@@ -35,7 +35,7 @@ sexp sexp_fixnum_to_bignum (sexp ctx, sexp a) {
   return res;
 }
 
-sexp sexp_make_integer (sexp ctx, sexp_lsint_t x) {
+sexp sexp_make_integer (sexp ctx, sexp_sint_t x) {
```
ashinn: We should preserve the ability to make integers from long ints.

okuoku: It won't allow declaring `sexp_lsint_t` as a `struct` on platforms such as Visual Studio Win64 -- it doesn't support larger integer types that would satisfy `sizeof(sexp_sint_t) < sizeof(sexp_lsint_t)`. As C99 does not require any integer types that would larger than platform's word size, requiring it would sacrifice portability.

When will we need `sexp_make_integer` from long ints? In case we should interface with C APIs, I think we should have separate C type variants, say `sexp_make_integer_from_long` or `sexp_make_integer_from_intptr` because we still need to cast an integer from C API to `sexp_(l)sint_t` which has platform-defined size.

Single `sexp_make_integer` API would require knowledge on width of long ints:

```c
uintptr_t res = some_foreign_api_returns_ptr();
sexp out = (sizeof(sexp_lsint_t) >= sizeof(long)) ?
  sexp_make_integer(ctx, res) : "blame we might not have BIGNUM here";
```

Just go straight with `sexp_make_integer` would implicitly truncate C API result -- fortunately, it won't occur on current chibi-scheme supported platforms though.

include/chibi/sexp.h
```diff
@@ -197,15 +197,20 @@ enum sexp_types {
...
-#define SEXP_PROC_NONE 0uL
-#define SEXP_PROC_VARIADIC 1uL
-#define SEXP_PROC_UNUSED_REST 2uL
-
 #ifdef _WIN32
+#if defined(_MSC_VER) && SEXP_64_BIT
+/* On SEXP_64_BIT, 128bits arithmetic is mandatory */
+#error Unsupported configuration
+#endif
+#ifdef SEXP_64_BIT
```
okuoku: `SEXP_64_BIT` shouldn't be evaluated with `#ifdef`. Addressed with https://github.com/ashinn/chibi-scheme/pull/438/commits/431bc38976ca39ed083fd0334dff9f1cd30d1d96

bignum.c
ashinn: sexp_make_integer is used frequently, notably from all FFI calls involving integer types, where we may not even trivially know whether it's a long if it's a typedef like offset_t.

sexp_lsint_t is just a long long int, which C99 guarantees support for.  It doesn't guarantee to be larger than a long, but the alternative sexp_make_integer_from_long doesn't improve the situation.

okuoku:
Hmm, regarding `sexp_make_integer` discussion, it looks like we would be better to stick with MSVC Win64 case. Consider MSVC Win64 has following integral types:

|type|width|
|----|------|
|`int`|32 bits|
|`long`|32 bits|
|`long long`|64 bits|

and the longest integral type is `long long`. On MinGW compiler, it has `int128_t` so we can specify `SEXP_64_BIT` but we cannot do that in straightforward way on MSVC.

I think we have three choices here:

1. Give up `SEXP_64_BIT` on MSVC Win64 and use different ABI between MinGW64.
2. Declare `sexp_lsint_t` as `struct` which has pair of 64bits and wrap `sexp_lsint_t` arithmetic in `bignum.c` -- to keep ABI compatibility between MinGW, `sexp_make_integer` should take C type
3. Give up `SEXP_64_BIT` on MinGW64 either

okuoku:
Rebased, added DL support and Appveyor.yml for MinGW32/64.

* Makefile: Moved some libraries to `CHIBI_POSIX_COMPILED_LIBS` and make it disable with `EXCLUDE_POSIX_LIBS`
* `sexp_allocated_bytes`: Should have `SEXP_API` for a library
* gc_heap.c `load_image_fn`: Stubbed out on Win32
* `sexp_make_flonum`: Needs `SEXP_API`

FIXME: It seems x64 has some error but it passed on Appveyor..

okuoku:
Reintroduced `msys` platform with `PLATFORM=msys`. I'm not sure we'd have any way to choose whether or not we're going to compile with MSYS's posix emulation layer..

To summarize build configurations that would supported by MSYS environment in this pull request:

- Native Win32: `uname` command says `Msys` and the Makefile automagically choose `windows` platform. Manually specifing `EXCLUDE_POSIX_LIBS=1` is required.
- Native Win64: `uname` command says `Msys` and the Makefile automagically choose `windows` platform. Manually specifing `EXCLUDE_POSIX_LIBS=1` is required.
- MSYS with POSIX emulation layer(fork of Cygwin): Manually specify `PLATFORM=msys`

On Win32 and MSYS, r7rs-test passes 100%. On Win64, `ci` related does not pass for now.

`EXCLUDE_POSIX_LIBS` is only required because `Makefile.detect` included too late.

okuoku:
Rebased and squashed commits. I believe this is the final ones for this PR.

* Postpone `sexp_make_integer` discussion
* Fix warnings

ashinn:
Thanks for all your help with windows support!

PR #439, opened 2017-11-05
RFC: Win32 feature identifier / library changes

okuoku:
This one is just a RFC use with #438 .

Historically, Chibi-scheme used `mingw` as platform feature identifier but I think it is not too correct because use of MSYS emulation layer is essentially Cygwin.. I'd recommend change it to `windows` as described in R7RS small.

I haven't tested at all with Makefile.detect.. I use CMake instead https://github.com/okuoku/chibi-scheme-cmake

- `(chibi filesystem)`: Perhaps we'd need more high-level subset which covers Win32 / Emscripten. The patch just stubs out most of the library.
- SRFI-27: Just uses `rand()` which is not recommended.
- SRFI-144: for `lgamma_r`, it is not available because it is not in C standard. I think we can copy it from Ruby https://github.com/ruby/ruby/blob/trunk/missing/lgamma_r.c

### DIFF
--- a/Makefile.detect
+++ b/Makefile.detect
@@ -20,7 +20,7 @@ ifeq ($(shell uname),DragonFly)
 PLATFORM=bsd
 else
 ifeq ($(shell uname -o),Msys)
-PLATFORM=mingw
+PLATFORM=windows
 SOLIBDIR = $(BINDIR)
 DIFFOPTS = -b
 else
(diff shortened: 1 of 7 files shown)


ashinn:
Thanks!  Let me verify there are no regressions for non-windows, and maybe find someone to verify mingw.

okuoku:
@ashinn : How should we go with `lgamma_r` ? Copy it from Ruby and embed into somewhere, or, implement it in Scheme-side?

ashinn:
For lgamma_r either should be fine.  There's a portable Scheme version in the SRFI reference implementation.

PR #440, opened 2017-11-08
Add mkdir and install commands for SRFI 117 to Makefile.

chaw:
'make install' was giving me an error about SRFI 117 not being found.  Based on a quick look at the Makefile, I think the attached changes make sense; they got rid of the error

Regards,

-chaw


ashinn:
Looks good, thanks!

PR #446, opened 2017-12-13
Win32: Wrap-up

okuoku:
Wrap-up Win32 port.

* `Makefile`: Autodetect MinGW platform using gmake-specific `$(OS)`. Now `make` will build against native Windows API by default; `EXCLUDE_POSIX_LIBS` is not required.
* `sexp.c`, `(chibi disasm)`, `port.c` and SRFI-98: Remove warnings on Win32.
* `README-win32.md`: New file to document current status and build instructions for Win32 port.
* Import public domain `lgamma_r` implementation from Ruby.
* `(chibi win32 process-win32)`: New library to implement `exit` on Win32, exported as part of `(scheme process-context)`
* `(chibi filesystem)`: Now Win32 implementation of `opendir` etc are included as a new separate file.


ashinn:
Thanks, lgtm!

PR #449, opened 2017-12-30
Several changes of syntax-rules in init-7.scm

Hamayama:
Hello.
I tried to use srfi-149 in Gauche and found some issues.
I changed 'syntax-rules' in init-7.scm as follows for feedback.

1. Add 'bad ellipsis' error check.
   Sorry, test case 'bad ellipsis' is now commented out
   because I coudn't handle error properly.

2. Literal has priority to ellipsis (R7RS 4.3.2).
   ( https://srfi-email.schemers.org/srfi-148/msg/6115633 )
   I added 'ellipsis-mark?' procedure.

3. Remove syntax-rules/aux.
   If 'ellipsis-specified?' is true, I use 'eq?' in 'ellipsis-mark?' ( same as literals ).
   If 'ellipsis-specified?' is false, I use 'compare' in 'ellipsis-mark?' (same as underscore ).
   By these changes, I considered that the binding of ellipsis and syntax-rules/aux could be removed.
   If this is wrong consideration, please tell me ...

### DIFF
--- a/lib/init-7.scm
+++ b/lib/init-7.scm
@@ -728,18 +728,22 @@
            _let (list (list v x))
            (cond
             ((identifier? p)
-             (if (memq p lits)
-                 (list _and
-                       (list _compare v (list _rename (list _quote p)))
-                       (k vars))
-                 (if (compare p _underscore)
-                     (k vars)
-                     (list _let (list (list p v)) (k (cons (cons p dim) vars))))))
+             (cond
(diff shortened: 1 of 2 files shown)


ashinn:
Thanks!  Sorry for the delay on this.  I think your change is correct, it is basically saying that the specified ellipsis should be compared with bound-identifier=?, the same as for literals.

PR #452, opened 2018-01-13
patch to compile on 9front

joe9:


ashinn:
Thanks!

okuoku:
This patch breaks C environments other than with APE, without `SEXP_USE_INTTYPES` because of `<ape/limits.h>`.

ashinn:
Sorry, there wasn't enough context in the diff and I thought it was a plan9 only section.  Fixed.

PR #462, opened 2018-03-29
Check for valid "rest" parameters in parameter lists

katterjohn:
Checks for invalid parameter names and duplicate parameters were being performed on parameter lists, but these checks were not considering any rest parameters.  So the behavior before this is that 

`(lambda (x x) x)` gives an error
`(lambda (x 0) x)` gives an error
`(lambda (x #t) x)` gives an error

but

`((lambda (x . x) x) 'foo 'bar)`  yields `foo`
`((lambda (x . 0) x) 'foo 'bar)` yields `foo`
`((lambda (x . #t) x) 'foo 'bar)` yields `foo`

and

`((lambda 0 'foo))` yields `foo`
`((lambda #t 'foo))` yields `foo`

And there are plenty of other examples.  Now these latter cases all produce errors.


ashinn:
Nice catch, thanks!

PR #463, opened 2018-03-31
Avoid placing duplicate feature identifiers in the features list

katterjohn:
In my particular case `bsd` was present twice on my OpenBSD and FreeBSD boxes because `sexp_platform` is `"bsd"` and `SEXP_BSD` is `1`:

`(chibi r7rs ratios complex threads full-unicode modules dynamic-loading bsd bsd little-endian)`

### DIFF
--- a/sexp.c
+++ b/sexp.c
@@ -438,7 +438,7 @@ static const char* sexp_initial_features[] = {
 void sexp_init_context_globals (sexp ctx) {
   const char** features;
   int i, endianess_check = 1;
-  sexp type, *vec, print=NULL;
+  sexp feature, type, *vec, print=NULL;
   sexp_context_globals(ctx)
     = sexp_make_vector(ctx, sexp_make_fixnum(SEXP_G_NUM_GLOBALS), SEXP_VOID);
 #if ! SEXP_USE_GLOBAL_SYMBOLS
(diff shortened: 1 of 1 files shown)


sexp.c
```diff
@@ -475,8 +475,11 @@ void sexp_init_context_globals (sexp ctx) {
   sexp_push(ctx, sexp_global(ctx, SEXP_G_FEATURES), SEXP_FALSE);
   sexp_car(sexp_global(ctx, SEXP_G_FEATURES)) = sexp_intern(ctx, (*(unsigned char*) &endianess_check) ? "little-endian" : "big-endian", -1);
   for (features=sexp_initial_features; *features; features++) {
-    sexp_push(ctx, sexp_global(ctx, SEXP_G_FEATURES), SEXP_FALSE);
-    sexp_car(sexp_global(ctx, SEXP_G_FEATURES)) = sexp_intern(ctx, *features, -1);
+    feature = sexp_intern(ctx, *features, -1);
+    if (sexp_not(sexp_memq(ctx, feature, sexp_global(ctx, SEXP_G_FEATURES)))) {
+      sexp_push(ctx, sexp_global(ctx, SEXP_G_FEATURES), SEXP_FALSE);
+      sexp_car(sexp_global(ctx, SEXP_G_FEATURES)) = feature;
```
ashinn: Either make feature a gc var, or call sexp_intern again here.

Note currently we never gc symbols so this isn't strictly needed, but that may change.

katterjohn: I just pushed another commit to (hopefully) take care of this.  Please let me know if this isn't the correct way to go about fixing this problem.  Thanks!

ashinn: lgtm, thanks!

PR #469, opened 2018-04-15
Raise error if apply is called with invalid args.

### LINKED ISSUE #468: apply is too lax
Currently, chibi allows some invalid calls to `apply`.

The R7RS standard indicates that the final argument, `args`, is required and must be a list:

```
(apply proc arg1 ... args)

The apply procedure calls proc with the elements of the list
(append (list arg1 ... ) args) as the actual arguments.
```

Interestingly, I think this is the only time a `...` appears in a non-terminating position in a procedure signature anywhere in the R7RS small standard. In this case the `...` applies to `args1`, which means that `args` is required. It is implied by `(append (list arg1 ... ) args)` that `args` must be a list. Other Scheme implementations require `args` to be provided, and to be a list.

Based on that description, there are two situations where chibi should signal an error, but currently does not.

### 1. Missing args list

First, `(apply proc)` should signal an error that not enough arguments were given. But currently `apply` is defined so that  behaves the same as `(proc)`:

https://github.com/ashinn/chibi-scheme/blob/10759e8bdb62d4e596a2631271e586ff6d1d7661/lib/init-7.scm#L46-L51

I think the solution to this is just to change line 48 to raise a "too few arguments" exception instead of calling `(proc)`.

### 2. Non-list final argument

The second situation is if the final argument is not a list, such as `(apply list 'a 'b)`. It should signal a type error, but instead it silently ignores the final argument. `(equal? (apply list 'a 'b) (list 'a))`

I'm not sure what is the cause of that exact behavior. Looking at the code, I think `sexp_append2` assumes the second argument is a list, so it builds an improper list by consing. And probably, somewhere deep in the implementation of `apply1`, I guess something loops until it finds a cdr which is not a pair, and assumes that it found the terminating empty list, so the value in the final cdr is ignored.

Whatever the cause is, a simple solution would be add a `(list? (car lol))` check in the lambda inside `apply`, and raise a type exception if it is not. The downside is that this check would impact performance of every call to `apply`.

If these solutions are acceptable to you, I can submit a pull request.

jacius:
Fixes #468

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -1174,10 +1174,18 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
     bc = sexp_procedure_code(self);
     cp = sexp_procedure_vars(self);
     ip = (sexp_bytecode_data(bc)+sexp_unbox_fixnum(stack[fp+1])) - sizeof(sexp);
-    for (top=fp-j+i-1; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
-      stack[top] = sexp_car(tmp2);
-    top = fp+i-j+1;
-    fp = k;
+    {
+      int prev_top = top;
+      for (top=fp-j+i-1; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
+        stack[top] = sexp_car(tmp2);
+      top = fp+i-j+1;
+      fp = k;
+      /* if final cdr of tmp2 isn't null, then args list was improper */
+      if (! sexp_nullp(tmp2)) {
+        top = prev_top;
+        sexp_raise("apply: improper args list", sexp_list1(ctx, stack[prev_top-2]));
+      }
+    }
     goto make_call;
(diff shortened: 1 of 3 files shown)


vm.c
```diff
@@ -1174,10 +1174,18 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
     bc = sexp_procedure_code(self);
     cp = sexp_procedure_vars(self);
     ip = (sexp_bytecode_data(bc)+sexp_unbox_fixnum(stack[fp+1])) - sizeof(sexp);
-    for (top=fp-j+i-1; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
-      stack[top] = sexp_car(tmp2);
-    top = fp+i-j+1;
-    fp = k;
+    {
+      int prev_top = top;
```
ashinn: Can you move the declaration to the beginning of the method?  Chibi is written to work with very old C compilers.

Issue #456, opened 2018-02-15
Add ABI-affecting header configs to install.h


ashinn:
Currently if you choose a non-standard compilation option which affects the runtime (e.g. Boehm GC), C programs linked to chibi must be compiled with the same options manually.  It would be much less error-prone if we could determine these and append them to install.h.

PR #475, opened 2018-06-09
More portable bignums

tramboi:
Hello, I'm liking chibi-scheme very much and I'd love to use it on MSVC x64 (for instance, but there might be others).
The bignum stack relies on gcc 128bit arithmetics so this pull request provides a portable (albeit much slower) reference backend for other compilers.
Please review when you have some time and tell me if some stuff needs to be changed,
cheers, and thank you for this Scheme :)
Bertrand


include/chibi/sexp.h
```diff
@@ -890,13 +890,13 @@ SEXP_API int sexp_idp(sexp x);
 #define _or_integer_flonump(x)
 #endif
 
+#define sexp_make_integer(ctx, x) sexp_make_fixnum(x)
+#define sexp_make_unsigned_integer(ctx, x) sexp_make_fixnum(x)
 #if SEXP_USE_BIGNUMS
-SEXP_API sexp sexp_make_integer(sexp ctx, sexp_lsint_t x);
-SEXP_API sexp sexp_make_unsigned_integer(sexp ctx, sexp_luint_t x);
```
ashinn: These are public APIs used by the FFI so need to be preserved (the sexp_make_fixnum definition is not compatible).  It's ok to ifdef though, only using sexp_l[su]int_t when these are native integer types, and fall back to long and unsigned long otherwise.

ashinn:
Thanks for doing this!  It has been on my todo list for a long time.  The overall approach looks good, I need a little more time to review.

include/chibi/sexp.h
tramboi: Mmmh I'm not really confortable with all the API/ABI implications of what I've done, so please massage it in any way you see fit :)

tramboi:
No problem, it's a pleasure and I understand people who don't want to/can't tackle MSVC/ARMCC/whatever obscure toolchain. And the code looks very sane and portable, so i wasn't a PITA at all to jump into it, so thank you.

ashinn:
I added some changes, mostly making this configurable and testable and preserving the FFI behavior.

If you make test-libs you'll find an infinite loop in the numeric tests on the internal sexp_bignum_sqrt. I haven't traced through to see if this is a new bug you introduced or existing bug you uncovered. If you want to take a look go ahead, otherwise I will likely just merge this and investigate separately.

tramboi:
Thanks for integrating this !
I'll check this tonight if I can.


2018-06-18 17:37 GMT+02:00 Alex Shinn <notifications@github.com>:

tramboi:
Ok, the division algorithm is botched. I'll fix this.
Btw, I think the code expects at least 64bit so that's probably why 32 bit
builds are broken.

2018-06-18 17:38 GMT+02:00 Bertrand Augereau <bertrand.augereau@gmail.com>:

okuoku:
> Btw, I think the code expects at least 64bit so that's probably why 32 bit builds are broken.

We need to decide a truth table for the definitions first; my proposal is we should have two separate API definitions for FFI on Win64:

|*ABI Macro*|GNU UNIX32|GNU UNIX64|GNU Win32|GNU Win64|MS Win32|MS Win64|
|--------|:----:|:----:|:----:|:----:|:----:|:----:|
|Proposal: `SEXP_64_BIT`|0|1|0|1|0|1|
|Proposal: `SEXP_USE_CUSTOM_LONG_LONGS`|0|0|0|0|0|1|
|Current: `SEXP_64_BIT`|0|1|0|1|0|0|
|Current: `SEXP_USE_CUSTOM_LONG_LONGS`|1|0|1|0|1|1|

Current = `#define SEXP_USE_CUSTOM_LONG_LONGS ((!SEXP_64_BIT) || !defined(__GNUC__))`

(ABI macros would go to #456 )

My choice prefer performance over ABI compatibility between FFI modules compiled with GNU compiler or MS compiler. Thus, `SEXP_USE_CUSTOM_LONG_LONGS` definitions do not same for GNU Win64 and MS Win64. (ie. We cannot use GNU Win64 compiled FFI modules on MS Win64 compiled chibi-scheme executable, vice versa.)

Another choice to consider is using same ABI with both GNU/MS Win64. 

@ashinn : CMake build needs to be fixed anyway because it assumes `SEXP_64_BIT=0` on MS Win64. I can fix it once we'd decided use of these macros.

tramboi:
Let's say that the initial implementation only provides
SEXP_USE_CUSTOM_LONG_LONGS when SEXP_64_BIT && !defined(__GNUC__).
With a buggy division, for now :)
It's a better situation than before, and once we get this right, I
think we'll be able to provide SEXP_USE_CUSTOM_LONG_LONGS when
!SEXP_64_BIT.


2018-06-18 23:50 GMT+02:00 okuoku <notifications@github.com>:

tramboi:
Ok guys, I fixed (probably, didn't explore the 256 bits tbh) the division and I made SEXP_USE_CUSTOM_LONG_LONGS dependent on SEXP_64_BIT.
Rebased -i all of this for a cleaner commit.

ashinn:
Lgtm.  @okuoku, agreed, I think it's fine to fix the cmake build in a separate patch so am merging this now.

PR #476, opened 2018-06-10
(expt bignum -k) was equal to (expt bignum k)

tramboi:
So here's a fix.
By the way there's an issue too with fixnums, (expt 3 -1) is not exact and I guess it should be.
Cheers,
Bertrand


ashinn:
lgtm, thanks!

PR #478, opened 2018-06-20
Activate Win64 platform with MSVC

okuoku:
Activate Win64 build and CI.

- Allow `SEXP_64_BIT` on Win64
- cmake: Never override `SEXP_64_BIT` on build
- Update README-win32.md


ashinn:
Thanks!

PR #480, opened 2018-07-03
Fix syntax-rules on ellipsis escape templates

Hamayama:
Hello.
I found that ellipsis escape templates do not expand pattern variables.

```
(define-syntax elli-esc-1
  (syntax-rules ()
    ((_ x)
     (... 'x))))
(elli-esc-1 100) ; ==> x (should be 100)
```

I tried to fix it.

### DIFF
--- a/lib/init-7.scm
+++ b/lib/init-7.scm
@@ -820,13 +820,10 @@
             ((null? p) (list _and (list _null? v) (k vars)))
             (else (list _and (list _equal? v p) (k vars))))))))
     (define ellipsis-mark?
-      (if (if ellipsis-specified?
-              (memq ellipsis lits)
-              (any (lambda (x) (compare ellipsis x)) lits))
-          (lambda (x) #f)
-          (if ellipsis-specified?
-              (lambda (x) (eq? ellipsis x))
-              (lambda (x) (compare ellipsis x)))))
+      (let ((cmp (if ellipsis-specified? eq? compare)))
+        (if (any (lambda (x) (cmp ellipsis x)) lits)
+            (lambda (x) #f)
+            (lambda (x) (cmp ellipsis x)))))
     (define (ellipsis-escape? x) (and (pair? x) (ellipsis-mark? (car x))))
     (define (ellipsis? x)
       (and (pair? x) (pair? (cdr x)) (ellipsis-mark? (cadr x))))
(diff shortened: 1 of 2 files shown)


Hamayama:
About the following part:
https://github.com/ashinn/chibi-scheme/pull/480/files#diff-35aa17e7d91c9beb3a08a1ff1ecc8016R876
```
            (lp (if (pair? (cdr t))
                    (if (pair? (cddr t)) (cddr t) (cadr t))
                    (cdr t))
                dim
                #t))
```

I wonder it can be reduced as follows.
Is this wrong?

```
            (lp (if (pair? (cdr t)) (cadr t) (cdr t)) dim #t))
```

lib/init-7.scm
```diff
@@ -820,13 +820,10 @@
...
-              (memq ellipsis lits)
-              (any (lambda (x) (compare ellipsis x)) lits))
-          (lambda (x) #f)
-          (if ellipsis-specified?
-              (lambda (x) (eq? ellipsis x))
-              (lambda (x) (compare ellipsis x)))))
+      (let ((cmp (if ellipsis-specified? eq? compare)))
+        (if (any (lambda (x) (cmp ellipsis x)) lits)
+            (lambda (x) #f)
+            (lambda (x) (cmp ellipsis x)))))
```
ashinn: This is slower in the common case, I wouldn't bother with the refactoring you did here.

ashinn:
The test in question checking (cddr t) is to distinguish (... x) which should expand to x, from the non-standard extension (... x y) which should expand to (x y).

ashinn:
BTW, thanks for catching this!  Overall the change looks good.

Hamayama:
> This is slower in the common case, I wouldn't bother with the refactoring you did here.

I understood. I will revert this to the original code.


> The test in question checking (cddr t) is to distinguish (... x) which should expand to x, from the non-standard extension (... x y) which should expand to (x y).

Uhm, I understood but should it be as follows because (cddr '(... x y)) ==> (y) ?

```
            (lp (if (pair? (cdr t))
                    (if (pair? (cddr t)) (cdr t) (cadr t))
                    (cdr t))             ^^^^^^^
                dim
                #t))
```

ashinn:
Oh, yes, I think we can simplify to:

```scheme
(lp (if (and (pair? (cdr t)) (null? (cddr t))) (cadr t) (cdr t)) dim #t)
```

Hamayama:
I fixed. Thanks!

Issue #484, opened 2018-08-25
SRFI 22 vs. (command-line) in (scheme process-context)


omasanori:
With SRFI 22, scripts can access the command line arguments via the argument of `main` procedure. On the other hand, R7RS provides `command-line` procedure to accomplish the same purpose. However, the results of these two methods differs on Chibi (version 0.8), while they are identical on Gauche (version 0.9.6) and Sagittarius (version 0.9.3). I did not tested Foment since it does not support SRFI 22 according to README.

Although neither R7RS-small nor SRFI 22 specifies whether they should be identical or not, I guess that it would be better if there is one consistent behavior between implementations, for the sake of portability. What do you think?

Test code:
```scheme
#! /usr/bin/env scheme-r7rs
(import (scheme base)
        (scheme process-context)
        (scheme write))

(define (main args)
  (write args)
  (newline)
  (write (command-line))
  (newline)
  0)

;(cond-expand
;  (srfi-138
;   (main (command-line)))
;  (else))
```

The result:
```shell
$ sudo ln -sfv gosh /usr/local/bin/scheme-r7rs
'/usr/local/bin/scheme-r7rs' -> 'gosh'
$ ./test.scm test
("./test.scm" "test")
("./test.scm" "test")
$ sudo ln -sfv sagittarius /usr/local/bin/scheme-r7rs
'/usr/local/bin/scheme-r7rs' -> 'sagittarius'
$ ./test.scm test
("./test.scm" "test")
("./test.scm" "test")
$ sudo ln -sfv chibi-scheme /usr/local/bin/scheme-r7rs
'/usr/local/bin/scheme-r7rs' -> 'chibi-scheme'
$ ./test.scm test
("./test.scm" "test")
("scheme-r7rs" "./test.scm" "test")
```

ashinn:
The SRFI 22 behavior agrees between implementations and with the SRFI 22 docs and examples.

The R7RS command-line description is:

>     Returns the command line passed to the process as a list
>     of strings. The first string corresponds to the command
>     name, and is implementation-dependent. It is an error to
>     mutate any of these strings.

This is perhaps a little underspecified, but I believe the chibi behavior is a valid interpretation.  However, this causes a problem in that the same program will behave differently when compiled vs run as a script.  I'm not sure we can clarify this as an errata.

omasanori:
> This is perhaps a little underspecified, but I believe the chibi behavior is a valid interpretation.

[...]

PR #485, opened 2018-08-29
Exclude argv[0] from (command-line) when invoked as scheme-r7rs.

omasanori:
This PR removes `argv[0]` from `(command-line)` when it is invoked as `scheme-r7rs`.

Another approach is changing `if (i >= argc || main_symbol != NULL)` to `if (i >= argc)` so `argv[0]` will be removed whenever the `-r` option is given. I think it is also reasonable and clear solution but it might break compatibility with existing scripts relying on the current behavior.

Fixes https://github.com/ashinn/chibi-scheme/issues/484


ashinn:
I've made this change universal (even when not running as scheme-r7rs).

PR #498, opened 2018-11-25
Emsdk update

amirouche:


amirouche:
There is a problem at least with emsdk precompiled `sdk-upstream-37382-64bit` where `clang` is looked up in the wrong directory. In `emsdk/upstream/37382/` directory, there is file named  `emscripten_config` with a variable named `WASM_INSTALL`. That must be replaced with something like:

```python
WASM_INSTALL = '/home/amirouche/src/c/emsdk/upstream/37382/'
```

amirouche:
I just re-compiled the thing using emsdk from git it works...

PR #499, opened 2018-11-29
Add PROCEDURALLY for outputting procedure objects

edw:
Using `(chibi show)` or SRFI-159 is difficult for debugging due to the difficulty with which procedure objects are written.


ashinn:
This is what `written` is for:

`(show #t (written map) nl)`

In general, if you want to write an object of an arbitrary type you should use written.

edw:
Ah, thanks for the tip!

PR #500, opened 2018-12-01
Implement syntax-case

mnieper:
Overview
=======

This patch implements the `syntax-case` macro facility for Chibi so that it works seamlessly with the existing `er-macro-transformer` and `syntax-rules` macro facilities. The `syntax-case` primitives are exported by the  `(chibi syntax-case)` library.

Examples
=======

Syntax
---------

The `syntax` form works like `quote` but renames all identifiers (and closes them in the transformer syntactic environment). For example, the macro transformer

```scheme
(define-syntax foo
  (er-macro-transformer
    (lambda (expr rename compare)
      `(,(rename 'cons) ,(rename '+) ,(rename '*)))))
```

can now be written more clearly as

```scheme
(define-syntax foo
  (er-macro-transformer
    (lambda (expr rename compare)
      (syntax (cons + *)))))
```

or (by using the special reader syntax) as

```scheme
(define-syntax foo
  (er-macro-transformer
    (lambda (expr rename compare)
      #'(cons + *))))
```

Quasisyntax
----------------

If one wants to embed code supplied by the macro user, one can now use `quasisyntax`. So instead of

```scheme
(define-syntax simple-let
  (er-macro-transformer
    (lambda (expr rename compare)
      `((,(rename 'lambda) (,(list-ref expr 1)) . ,(list-tail expr 3)) ,(list-ref expr 2)))))
```

one can now write

```scheme
(define-syntax simple-let
  (er-macro-transformer
    (lambda (expr rename compare)
      #`((lambda (#,(list-ref expr 1)) . #,(list-tail expr 3)) #,(list-ref expr 2)))))
```

Rename
-----------

Obviously, the `rename` procedure is not needed when `syntax` or `quasisyntax` is used. And, in fact, `(rename (syntax-quote x))` is nothing but `(syntax x)`. The previous example can be written more briefly as:

```scheme
(define-syntax simple-let
  (lambda (expr)
    #`((lambda (#,(list-ref expr 1)) . #,(list-tail expr 3)) #,(list-ref expr 2)))))
```

The `compare` procedure is provided by the globally defined `free-identifier=?` procedure.

Pattern variables
----------------------

The `syntax` and `quasisyntax` forms can view certain identifiers as pattern variables that are automatically substituted without having to use `unsyntax`. Pattern variables are bound with `with-syntax`. The previous example simplifies to:

```scheme
(define-syntax simple-let
  (lambda (expr)
    (with-syntax ((var (list-ref expr 1)) (init (list-ref expr 2)) (body (list-tail expr 3)))
      #'((lambda (var) . body) init))))
```

Syntax-case
----------------

Finally, destructuring the input form can be done with `syntax-case`, which is nothing but a pattern matcher that binds pattern variables:

```scheme
(define-syntax simple-let
  (lambda (expr)
    (syntax-case expr ()
      ((_ var init . body) #'((lambda (var) . body) init)))))
```

Implementation notes
================

Most of the code is the Scheme implementation code of the `(chibi syntax-case)`. Here, the majority of the code implements the pattern matcher of `syntax-case` and the templating engine of `syntax` and `quasisyntax`, which is no more complicated than implementing `syntax-rules` and `quasiquote`.

A few lower-level changes have been necessary: While it is not directly linked to the `syntax-case` system, macro transformers in the R6RS take only one argument the expression. In order to make Chibi compatible, `define-syntax` and `let(rec)-syntax`'s transformer procedures now expect one element. The original versions are still accessible under the names `%define-syntax` and `%let(rec)-syntax`. Luckily, the `er-macro-transformer` and the `sc-macro-transformer` systems of Chibi are not incompatible to R6RS's requirements.

Chibi's two readers have been extended so that they now understand the special reader syntax for `syntax`, `quasisyntax`, `unsyntax` and `unsyntax-splicing`.

Macro values have received a fourth field (named `aux`). This is used to implement pattern variables. Pattern variables are macros with extra information stored in the `aux` field so that `syntax` can detect which identifiers are pattern variables and which are not.

A minor error in lib/chibi/ast.c had to be corrected.

All in all, implementing `syntax-case` for Chibi's expander was very straight-forward and no fundamental changes were necessary so that it has mostly become one library of Scheme code.

TODO
=====

`make-variable-transformer` and `identifier-syntax` haven't been implemented yet, but it won't be hard. With them and the above, `(rnrs syntax-case)` would be fully implemented.


lib/chibi/syntax-case.scm
```diff
@@ -0,0 +1,323 @@
...
+  (strip-syntactic-closures stx))
+
+(define (symbol->identifier id symbol)
+  (if (symbol? id)
+      symbol
+      ((syntactic-closure-rename id)
+       symbol)))
+
+;; TODO: Handle cycles in datum.
+(define (datum->syntax id datum)
```
ashinn: I think it's worth having datum->syntax in the core.

lib/chibi/syntax-case.scm
```diff
@@ -0,0 +1,323 @@
...
+		    (gen-template (cadr expr) '() ellipsis? level)))
+       out))))
+
+(define (syntax->datum stx)
+  (strip-syntactic-closures stx))
+
+(define (symbol->identifier id symbol)
+  (if (symbol? id)
+      symbol
+      ((syntactic-closure-rename id)
```
ashinn: This can be false with sc- and rsc- macros, right?

ashinn:
This is really cool!  Still going through it.

lib/chibi/syntax-case.scm
mnieper: Sounds good. I'm moving it back to init-7.scm and re-export it from `(chibi syntax-case)` (so that all relevant bindings can also be found in one place for programs that do not want to import the whole of `(chibi)`).

In the long run, one may think to move even `syntax` and `syntax-case` into the core. Not because they need to be exported by `(chibi)` but because the code needed to implement `syntax-rules` in the core is roughly the same code needed to implement `syntax-case` and `syntax` and we could avoid code-doubling.

By defining `syntax`/`quasisyntax` first, I was able to use it in subsequent macros (like `syntax-case`) making their code much easier to read and write (because I didn't have to use `rename` all over the place). Some macros in `init-7.scm` may also benefit from this.

lib/chibi/syntax-case.scm
mnieper: I don't see how to implement `datum->syntax` for sc- and rsc-macros. I'm not even sure how compatible they are to syntax-rules macros in the case of unhygienic macros that need `datum->syntax` on the syntax-rules/er-macro-transformer side.

Thus in `symbol->identifier` the identifier `id` (if it is not a symbol) has to be an identifier created by the renaming procedure of `er-macro-transformer` or of the form `#'x` (where `x` is a syntactic identifier), which is basically the same.

General identifiers created by `make-syntactic-closure` simply do not contain enough information about their syntactic environment. `symbol->identifier` could either raise an error in this case or return `symbol` unchanged.

mnieper:
> This is really cool! Still going through it.

Thanks! I'm glad that you like it. By writing this code I have learned that the syntax-case system can be viewed just as er-macro-transformer plus the two constituents of syntax-rules (namely the pattern matcher and the templating engine). (And er-macro-transformer together with just `syntax` gives a system of comparable power to ir macros.)

lib/srfi/99/records/syntactic.scm
```diff
@@ -10,11 +10,13 @@
             (procs (cddr expr))
             (make (car procs))
             (make-name (if (eq? make #t)
-                           (string->symbol (string-append "make-" name-str))
+                           (datum->syntax (car expr)
+					  (string->symbol (string-append "make-" name-str)))
```
Hamayama: I wonder `(car expr)` should be `(cadr expr)` for the compatibility to SRFI 99 reference implementation.
This changes the macro expansion result if define-record-type and record-name are inserted separately.

mnieper: Thanks for the comment.

The spec does not seem to be totally clear about it. I haven't taken a look at SRFI 99's reference implementation, so `(car expr)' was my reading of the spec. However, `(cadr expr)' makes at least as much sense and we shouldn't needlessly deviate from the reference implementation.

R6RS has a similar feature where the constructor and predicate name are derived from the record name. I have taken a look at Chez's implementation. It is also consistent with SRFI 99's reference implementation in that the lexical context is taken from the record name identifier and not from the lexical context of the macro call.

lib/chibi/syntax-case.scm
ashinn: How about we warn and create a new identifier with the `current-renamer`?

mnieper: What do you think of my idea (which I wrote about yesterday in Chibi's newsgroup) to attach the renamer not to the syntactic closure representing an identifier but to the environment (which is stored inside the syntactic closure)? In that case, all transformer environments would bundle their (non-trivial) renamer. For all other environments, the identity renamer would be the default.

ashinn:
Startup and image size look good - this is mostly pay-as-you-go for users of syntax-case.

mnieper:
Great! Will create a new pull request when I have improvements (for example, of taking renaming into account when interpreting the FREE-NAMES argument to `make-syntactic-closure`). And (as said above), merging the code in `(chibi syntax-case)` with the code for syntax-rules in init-7.scm makes sense in the long run in my opinion.

ashinn:
I wouldn't worry too much about merging syntax-rules if the syntax-case version requires notably more code in the core.

mnieper:
It shouldn't require notably more code. Most of the code (and all non-trivial code) in `(chibi syntax-case)` is the pattern matcher and the templating engine, which is exactly what has to be implemented for syntax-rules anyway.

PR #508, opened 2019-01-16
Fix tiny Makefile typo for SRFI 160 install.

chaw:
Trivial typo fix that I needed to make for 'make install' to work.


ashinn:
Thanks for the report and fix!  Merged in pr #509.

PR #530, opened 2019-03-13
Fix tests on Windows

Oxyd:
What this doesn't solve is the copious amount of warnings Visual Studio produces. The warnings are all for a good reason: The codebase doesn't seem to be able to decide whether it wants to represent things as `int` or `long` or `sexp_sint_t` or `sexp_uint_t` or `sexp_tag_t`, and the code is full of conversions from a larger type to a smaller one, or from unsigned to signed or vice-versa. This inevitably results in overflows like this: `max` overflowed to `0` when reading a `#u32(anything)`, and `(long)sexp_unbox_fixnum(obj)` overflowed to `4294967295` when `obj` represented a `-1`.

Would you be interested in me going over the codebase and fixing the warnings? This would mostly entail changing types of things to the correct types (such as from `long` to `sexp_sint_t`) where meaningful and adding explicit casts elsewhere.

### DIFF
--- a/sexp.c
+++ b/sexp.c
@@ -2280,7 +2280,7 @@ sexp sexp_write_one (sexp ctx, sexp obj, sexp out) {
       break;
     }
   } else if (sexp_fixnump(obj)) {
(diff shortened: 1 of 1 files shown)


ashinn:
Sure, I'm open to such patches but maybe do it incrementally to avoid a monster patch.

PR #538, opened 2019-04-25
Update TODO

kathe:
Requesting change of code License.


ashinn:
Chibi is already using a New BSD (3 clause) license in COPYING.

PR #551, opened 2019-07-24
Fix include-shared for process

v01dXYZ:
`chibi-genstatic` doesn't support expansion of `cond-expand` when the pattern is `(cond-expand (cond) (else (include-shared "module")))`


v01dXYZ:
*Purpose of the PR* By not expanding it, chibi-genstatic doesn't generate the C functions associated to this module, making a static linked chibi-scheme unworkable.

PR #554, opened 2019-08-02
Add features for the various BSDs.

apg:
This could and should be done cleaner. Works as a proof of concept at
this point in time, if it's an interesting patch to take on.


apg:
OK, per @johnwcowan's recommendation, I've expanded upon the feature detection instead of using `defined(...)` within sexp.c. Hopefully this is the approach you'd want to see.

PR #555, opened 2019-08-07
build on Solaris-like systems

richlowe:
This lets you build on illumos or Solaris.  I'm not sure if any of it is really done correctly.

We need to link to `libsocket` for `shutdown` and define `_POSIX_PTHREAD_SEMANTICS` to get the right versions of `asctime_r` and such.  Add a `PLATFORM` "solaris" to condition those.

time has no tm_zone or tm_gmtoff, so they are conditionalized and not present.  They could also be supplied in a Solaris-y manner, but I don't know how to arrange for the stub file to generate an arbitrary function.

### DIFF
--- a/Makefile.detect
+++ b/Makefile.detect
@@ -35,7 +35,11 @@ else
 ifeq ($(shell uname -o),GNU/Linux)
 PLATFORM=linux
 else
+ifeq ($(shell uname),SunOS)
+PLATFORM=solaris
+else
 PLATFORM=unix
+endif
 endif
 endif
 endif
@@ -70,6 +74,14 @@ EXE =
 CLIBFLAGS = -fPIC
 CLINKFLAGS = -shared
 LIBDL = 
+RLDFLAGS=-Wl,-R$(LIBDIR)
+else
+ifeq ($(PLATFORM),solaris)
+SO = .so
+EXE =
+CLIBFLAGS = -fPIC
+CLINKFLAGS = -shared
+LIBDL = -ldl
 RLDFLAGS=-Wl,-R$(LIBDIR)
 else
 ifeq ($(PLATFORM),windows)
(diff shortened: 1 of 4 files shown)


Makefile.detect
```diff
@@ -35,6 +35,9 @@ else
 ifeq ($(shell uname -o),GNU/Linux)
 PLATFORM=linux
 else
+ifeq ($(shell uname),SunOS)
+PLATFORM=solaris
+else
```
ashinn: Maybe keep the unix feature since solaris is a specific unix?

lib/chibi/time.stub
```diff
@@ -7,21 +7,37 @@
...
+    (int     tm_min    time-minute)
+    (int     tm_hour   time-hour)
+    (int     tm_mday   time-day)
+    (int     tm_mon    time-month)
+    (int     tm_year   time-year)
+    (int     tm_wday   time-day-of-week)
+    (int     tm_yday   time-day-of-year)
+    (int     tm_isdst  time-dst?)
+    (string  tm_zone   time-timezone-name)
+    (int     tm_gmtoff time-offset))))
```
ashinn: This is where CL's #+ comes in handy.

Makefile.detect
richlowe: And then check `uname` to guard `-lsocket`?  I'm not sure if it's appropriate on other things `unix` would apply to, since it's basically the default case.

ashinn: Sorry, not here, I meant in sexp.h at the Scheme level define both solaris and unix.

richlowe: Well, it's definitely better not to put "solaris" in there twice.  I've updated not to do that, at least.

I'm never sure what people think "unix" would apply to.  It seemed weird to me for "solaris" to imply "unix" but "bsd" not to, or "linux" not to, so I didn't do that bit.

Would
```c
#if !defined(_WIN32)
  "unix"
#endif
```
Be closer to the right thing? (until a VMS user comes along, I suppose...)

ashinn: Ah, nevermind, I thought unix was already defined there.  Looks like we set bsd along with the specific bsd but otherwise don't have multiple platform features.

The pr looks good.

PR #556, opened 2019-08-16
Fix NetBSD, DragonFly BSD and Darwin feature detection

katterjohn:
On NetBSD systems the `netbsd` feature identifier was not actually provided because the macro `SEXP_NETBSD` was defined twice and the second value was 0.

On DragonFly systems the `dragonfly` feature identifier was not actually provided because the macro `SEXP_DRAGONFLY` was defined twice and the second value was 0.  Also on DragonFly systems the `darwin` feature identifier was incorrectly provided because `SEXP_DARWIN` had a value of 1 instead of 0.

Tested on OpenBSD, NetBSD and DragonFly BSD.  The corresponding defines that are already present for Darwin and FreeBSD look correct.

### DIFF
--- a/include/chibi/features.h
+++ b/include/chibi/features.h
@@ -323,14 +323,12 @@
 #define SEXP_NETBSD 1
 #define SEXP_DRAGONFLY 0
 #define SEXP_OPENBSD 0
-#define SEXP_NETBSD 0
 #elif defined(__DragonFly__)
-#define SEXP_DARWIN 1
+#define SEXP_DARWIN 0
 #define SEXP_FREEBSD 0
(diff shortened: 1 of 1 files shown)


lassik:
Thank you to the people working on BSD support.

Is there a reason the `SEXP_*BSD` macros are defined as 0 or 1 instead of simply being defined or undefined and using `#ifdef` to check?

katterjohn:
Well, it's not just these macro definitions and corresponding checks.  In `include/chibi/features.h` other macros are done the same way, and `#if` is used far more than `#ifdef` in Chibi's code.  I avoided switching to `#ifdef` here for consistency.

Issue #558, opened 2019-08-21
compile error in (chibi pty) on macOS


mbamac:
(chibi pty) do not compile on macOS Mojave (and High Sierra) with error:

cc  -dynamiclib -DSEXP_USE_NTPGETTIME -DSEXP_USE_INTTYPES -Iinclude  -Wall -g -g3 -O3   -o lib/chibi/pty.dylib lib/chibi/pty.c -L.   -lchibi-scheme -lutil
lib/chibi/pty.c:5:10: fatal error: 'pty.h' file not found
#include <pty.h>
         ^~~~~~~
1 error generated.
make: *** [lib/chibi/pty.dylib] Error 1

lassik:
Try changing `(c-system-include "pty.h")` to `(c-system-include "util.h")` in `lib/chibi/pty.stub`

iizukak:
@lassik That's works for me. thanks.

PR #559, opened 2019-08-22
(chibi pty): fix compilation errors on the BSDs

katterjohn:
Some changes are necessary to fix compilation on the BSDs:

- Include `util.h` instead of `pty.h` on OpenBSD and NetBSD.

- Include `libutil.h` instead of `pty.h` on FreeBSD and DragonFly BSD.  DragonFly actually has both `util.h` and `libutil.h`, but the man pages say `libutil.h` so that's what I went with.

- Include `utmp.h` only on non-BSD systems (FreeBSD does not have that header and none of these BSDs require it).  I don't know if we can get away with including it only on Linux, so I stuck to non-BSD.

I tested on all four of these BSDs.

See also #558

### DIFF
--- a/lib/chibi/pty.stub
+++ b/lib/chibi/pty.stub
@@ -1,8 +1,10 @@
 
 (cond-expand
- (macosx (c-system-include "util.h"))
+ ((or macosx openbsd netbsd) (c-system-include "util.h"))
+ ((or freebsd dragonfly) (c-system-include "libutil.h"))
  (else (c-system-include "pty.h")))
-(c-system-include "utmp.h")
+(cond-expand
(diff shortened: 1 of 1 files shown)


katterjohn:
The current AppVeyor build failures seem to be unrelated to my commit.  They seem to fail because `pty.h` is also unavailable on some of those setups, but I'm not sure what to do on those.

It's failing with MinGW and MSVC using cmake, although it's succeeding with MinGW using MSYS2.  I don't have any of those systems to play with, so I haven't investigated any further.

ashinn:
This isn't going to work on windows, that's expected.

PR #560, opened 2019-08-23
(chibi pty): Disable in CMake build

okuoku:
Disable `(chibi pty)` build with CMake.

Unfortunately, AppVeyor doesn't have any support to "watch" CI on another account. Feel free call/assign me anytime if you have any issue on AppVeyor build...

FWIW, [Windows will support PTY soon](https://devblogs.microsoft.com/commandline/windows-command-line-introducing-the-windows-pseudo-console-conpty/) but I don't think it's worth to add it as Windows does not support the other essential POSIX feature -- `fork` .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -68,7 +68,8 @@ if(WIN32)
         lib/chibi/process.sld
         lib/chibi/stty.sld
         lib/chibi/system.sld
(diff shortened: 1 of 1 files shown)


ashinn:
Interesting... I'd be happy to support only the high-level `call-with-pty-process-io` if we could make it work on Windows as well.

PR #563, opened 2019-08-29
Change snow-chibi to work with Chicken 5

fisherro:
The method snow-chibi used to get the version of Chicken installed
doesn't work with Chicken 5. Adding "-R chicken.platform" would make it
work with Chicken 5, but then it wouldn't work with Chicken 4.

In both Chicken 4 & 5, however, csi has a -release option that will just
give the version number. So, I've changed the Chicken version detection
in snow-chibi to use this option.

Testing with...

	./chibi-scheme tools/snow-chibi.scm implementations

...then seemed to work with both Chicken 4 & 5.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -1287,6 +1287,16 @@
       (string->number (process->string '(csi -p "(##sys#fudge 42)")))
       8))
 
+(define (get-chicken-repo-path)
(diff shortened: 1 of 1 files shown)


fisherro:
Hmm. Well, this fixes the version check, but it is still giving me an error when I try to install a package.

fisherro:
I pushed a fix for finding the Chicken 5 repository path.

ashinn:
Thanks!

PR #567, opened 2019-09-10
(chibi net): sockaddr-port: return the port number in host byte order

katterjohn:
I expected the following to give me 22, but in this case it was being given to me in the wrong byte order:

```
> (sockaddr-port (address-info-address (get-address-info "localhost" 22)))
5632
```


katterjohn:
For comparison, I just tested using the socket egg on Chicken and it gives me 22 as I expected:

```
#;2> (sockaddr-port (inet-address "127.0.0.1" 22))
22
```

PR #568, opened 2019-09-10
(chibi net): sockaddr-name: fix support for IPv6 addresses

katterjohn:
The buffer was too small to hold all IPv6 addresses in string form.  `INET6_ADDRSTRLEN` is the standard constant giving the required length to hold these addresses.

### DIFF
--- a/lib/chibi/accept.c
+++ b/lib/chibi/accept.c
@@ -92,7 +92,7 @@ sexp sexp_listen (sexp ctx, sexp self, sexp fileno, sexp backlog) {
 /* Additional utilities. */
 
 sexp sexp_sockaddr_name (sexp ctx, sexp self, struct sockaddr* addr) {
-  char buf[24];
+  char buf[INET6_ADDRSTRLEN];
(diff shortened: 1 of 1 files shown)


lassik:
Excellent work mapping all of these out Kris. On my system, `INET6_ADDRSTRLEN` is 46. I guess that's 32 for the hex digits and 7 for the colon separators. Where do the remaining characters go?

katterjohn:
Hi Lassi.

POSIX actually defines it to be 46. It needs to be 46 to allow for writing an IPv4-mapped address in an uncompressed form.

The compressed form would be something like `::123.123.123.123`, while the uncompressed form would be like `0000:0000:0000:0000:0000:0000:123.123.123.123`.

So that is 45 bytes, and the NUL makes 46.

katterjohn:
Oh, it occurs to me now that I should perhaps mention this for anyone looking at the discussion but not the diff itself: this commit is not specifically for fixing the IPv4 compatibility stuff I previously mentioned.  The original buffer size was only 24, so there were plenty of "regular" IPv6 addresses that were too large to fit.

PR #575, opened 2019-09-23
(chibi net): make-listener-socket: close the socket on error

katterjohn:
A socket was leaked in the case where setting `socket-opt/reuseaddr` failed.  (The socket was already being closed in the other cases where `bind` or `listen` failed.)


ashinn:
Thanks for catching this.

PR #580, opened 2019-10-06
Fix incorrect and syntax

yamacir-kit:
According to R7RS 4.2.1, the and syntax with no arguments is correct to return true.


ashinn:
This is already working as intended, you can verify `(and) => #t` in chibi.

The result of a cond clause with no expression is the value of the test, which in this case is `#t`.

yamacir-kit:
I read R7RS again and realized that I was wrong.
Thank you for your kind reply.

PR #581, opened 2019-10-19
Add (get-stack-trace) special form to (chibi).

lockywolf:
Not sure if this is correct implementation, and even less sure if the garbage collection magic is correct, but this seem to actually make a string out of stack trace.


chibi-scheme.SlackBuild
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash -x
```
ashinn, line 1: Sorry, I lost track of this PR.  What's the SlackBuild for?

lockywolf:
I think this PR is outdated, I need to make a review myself of which changes would be beneficial to chibi.

Issue #584, opened 2019-10-23
SXML no option for safe text


ekaitz-zarraga:
Hi,
It would be nice to include a mark to avoid sxml content to be escaped.
That way different ways to create the XML can be mixed.

I'd like to make this patch myself but I'd like to know first if this is useful and I'd need some guide to find how to write it down.

Thanks!

PR #585, opened 2019-10-23
[WIP] Add option to ignore escaping

ekaitz-zarraga:
*I'd like to get feedback about this functionality.*

The idea is to add the chance to set a field in the SXML that is renders the string as it is as proposed in #584

### DIFF
--- a/lib/chibi/sxml.scm
+++ b/lib/chibi/sxml.scm
@@ -93,6 +93,9 @@
                   (display (html-tag->string tag (cdar rest)) out)
                   (for-each lp (cdr rest))
                   (display "</" out) (display tag out) (display ">" out))
+                 ((and (eq? '@raw tag)
+                       (string? (car rest)))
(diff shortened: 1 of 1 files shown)


lib/chibi/sxml.scm
```diff
@@ -93,6 +93,9 @@
                   (display (html-tag->string tag (cdar rest)) out)
                   (for-each lp (cdr rest))
                   (display "</" out) (display tag out) (display ">" out))
+                 ((and (eq? 'safe-string tag)
```
ashinn: I think "raw" or "unescaped" would be a better term - this is explicitly an unsafe operation!

It's also probably better to avoid collision with valid tags, maybe `(@unescaped <str>)`.

ekaitz-zarraga: Updated and chose @raw as you said.

PR #587, opened 2019-10-28
[wip] add srfi-168 with dependencies

amirouche:
some help


amirouche:
I plan to add the following SRFIs:

- 146 as known as `(scheme mapping)` & `(scheme mapping hash)`
- 158 as known as `(scheme generator)`
- 167 in-memory implementation as known as `okvs`
- 168 as known as `nstore`

I also would like to add a sqlite lsm based implementation of SRFI-167 but that would add an external dependency to chibi... like sqlite the source can be embedded in the repository. 

Let me know what you think.

ashinn:
(srfi 146) is already included, as is (scheme generator).

I will include anything in the large language.  Anything else needs a rationale, such as either being incredibly useful or not possible to implement portably.

(chibi sqlite3) is already available on snow-fort.org.

amirouche:
> (srfi 146) is already included, as is (scheme generator).

Ok.

> I will include anything in the large language. Anything else needs a rationale, such as either being incredibly useful or not possible to implement portably.

It is not the case of srfi-168.

I leave the PR open for the time being.

amirouche:
No need to keep it open.

PR #588, opened 2019-10-28
WIP: Suggest commands with levensthein distance

ekaitz-zarraga:
Work in progress of a suggestion system for chibi.app.

I think I need some help and suggestions

### DIFF
--- a/lib/chibi/app.scm
+++ b/lib/chibi/app.scm
@@ -332,6 +332,36 @@
 ;;> all prefixed by \var{prefix}.  The original \var{spec} is used for
 ;;> \scheme{app-help}.
 
+
+(define (levenshtein s t)
+  (define (%levenshtein s sl t tl)
+    (cond ((zero? sl) tl)
+          ((zero? tl) sl)
+          (else
+            (min (+ (%levenshtein (cdr s) (- sl 1) t tl) 1)
+                 (+ (%levenshtein s sl (cdr t) (- tl 1)) 1)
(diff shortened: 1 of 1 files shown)


lib/chibi/app.scm
```diff
@@ -332,6 +332,36 @@
...
+          ((zero? tl) sl)
+          (else
+            (min (+ (%levenshtein (cdr s) (- sl 1) t tl) 1)
+                 (+ (%levenshtein s sl (cdr t) (- tl 1)) 1)
+                 (+ (%levenshtein (cdr s) (- sl 1) (cdr t) (- tl 1))
+                    (if (char=? (car s) (car t)) 0 1))))))
+  (%levenshtein (string->list s)
+                (string-length s)
+                (string->list t)
+                (string-length t)))
```
amirouche: That implementation of levenshtein distance is slow (maybe too slow).

If it works for you ok. Otherwise try to adapt this implementation: https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Levenshtein_distance#C

amirouche: Oh! I should I read the whole PR. I think the implementation of levenshtein distance you picked is good enough.

ashinn: We actually want a levenshtein automata, but pushed a smarter impl for now in `(chibi edit-distance)`.

ekaitz-zarraga: Great! I'll remove it.

PR #591, opened 2019-11-07
(chibi show): Make ANSI escape codes work with col state variable

krzygorz:
The `col` state variable is increased by `output-default` in `(chibi show base)` every time it receives a chunk of formatted text. This is fine as long as `output-default` never receives a fragment of an ANSI escape code or Unicode char. Unfortunately, `(chibi show color)` uses `each` to construct escape sequences breaking this assumption.

For example,
```scheme
(show #f (as-unicode (as-red "1234567") (fn ((col)) (each " col: " col))))
```
evaluates to `"\x1b;[31m1234567\x1b;[0m col: 16"` rather than `"\x1b;[31m1234567\x1b;[0m col: 7"`.

This PR changes the `ansi-escape` procedure of `(chibi show color)` to return a single string rather than a  composite formatter.

### DIFF
--- a/lib/chibi/show-test.sld
+++ b/lib/chibi/show-test.sld
@@ -739,6 +739,10 @@ def | 6
           (show #f (with ((pad-char #\〜)) (padded/both 5 "日本語"))))
       (test "日本語"
             (show #f (as-unicode (with ((pad-char #\〜)) (padded/both 5 "日本語")))))
+      (test "\x1b;[31m1234567\x1b;[0m col: 7"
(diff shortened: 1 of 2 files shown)


krzygorz:
Another possible fix would be to leave `ansi-escape` as it is and modify `output-default` to use a new state variable that keeps track of partially written multi-byte characters. However, this approach would likely make `as-unicode` not equivalent to `(with ((string-width unicode-terminal-width)) fmt ...)`.

PR #596, opened 2019-12-26
Add (chibi apropos) module

lassik:
Here's are Common Lisp -style `apropos` and `apropos-list` procedures for finding procedures in the REPL.

The use of `eval` is not ideal but I used it since `env-ref` is not available from Scheme.

### DIFF
--- a/lib/chibi/apropos.scm
+++ b/lib/chibi/apropos.scm
@@ -1,0 +1,33 @@
+(define (list-bindings env)
+  (let parents ((env env) (binds '()))
+    (if (not env) binds
+        (let symbols ((syms (env-exports env)) (binds binds))
+          (if (null? syms) (parents (env-parent env) binds)
+              (symbols (cdr syms) (if (assv (car syms) binds) binds
+                                      (cons (cons (car syms) env)
+                                            binds))))))))
+
+(define (apropos-list-bindings query)
+  (cond ((symbol? query) (set! query (symbol->string query)))
+        ((not (string? query))
+         (error "Apropos query must be a symbol or a string")))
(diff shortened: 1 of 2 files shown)


lib/chibi/apropos.scm
```diff
@@ -0,0 +1,36 @@
...
+        (lambda (a b) (string<? (symbol->string (car a))
+                                (symbol->string (car b))))))
+
+(define (apropos-list query) (map car (apropos-list-bindings query)))
+
+(define (apropos-prefix sym env)
+  (let ((p "procedure  ")
+        (s "syntax     ")
+        (v "variable   "))
+    (call/cc (lambda (bail)
```
ashinn: No need for call/cc, you can just return from the exception handler normally.

lassik: How would you write that? `(with-exception-handler (lambda (_) s) (lambda () (eval sym env)))` causes `ERROR: exception handler returned` when `sym` is a macro

ashinn: Oh, you would need guard, if you want the lighter with-exception-handler you do need call/cc.

lassik: OK, right. I force-pushed a new commit changing it to use `guard` instead. It's still a bit of a hack in the absence of an `env-ref` in Scheme - a full `eval` is not really needed as all bindings should point to self-evaluating objects.

lassik:
"Screenshot":

```

> (import (chibi apropos))
> (apropos "/")
procedure  /
procedure  call/cc
procedure  floor/
syntax     read/ss
procedure  truncate/
syntax     write/ss
```

Issue #531, opened 2019-03-29
SRFI 125: Hash functions


mnieper:
Some SRFI 125 functions call SRFI 128 hash functions with two arguments, which is an error.

ar-nelson:
Seconded, this bug is preventing me from using hash tables in Chibi. It can be replicated with a small Scheme program, `problem.scm`:

```scheme
(import (scheme base) (srfi 125) (srfi 128))
(define ht
  (make-hash-table (make-vector-comparator (make-default-comparator)
                                           vector?
                                           vector-length
                                           vector-ref)))
(hash-table-set! ht #(1 2) 'foo)
```

This program crashes with a "too many args" error:

```
$ chibi-scheme problem.scm
ERROR in final-resumer: too many args
    #<procedure #f>
    2
```

It seems like the problem is in SRFI 128, not SRFI 125. Chibi's SRFI 125 is implemented on top of SRFI 69, which expects two-argument hash functions. The SRFI 128 document [specifically warns about this possibility](https://srfi.schemers.org/srfi-128/srfi-128.html#Hashfunctions):

> ...if programmers wish their hash functions to be backward compatible with the reference implementation of SRFI 69, they are advised to write their hash functions to accept a second argument and ignore it.

The ideal solution would be to modify `make-comparator` in SRFI 128 to wrap the hash function in a lambda that takes a second argument and ignores it. I'll submit a PR that does this.

vuolen:
Should this be closed?

ashinn:
Yes.

PR #604, opened 2020-01-11
Fix make-comparator hash function arity in SRFI 128

ar-nelson:
This resolves issue #531: SRFI 125 functions would sometimes crash when using SRFI 128 comparators with single-argument hash functions, because SRFI 125 is based on SRFI 69, which expects two-argument hash functions.

This PR resolves the issue by changing `make-comparator` to wrap single-argument hash functions in a lambda that takes and ignores a second argument. It uses `(chibi ast)` to detect functions that could accept two arguments, so that it doesn't unnecessarily wrap them and remove any functionality from the second argument.

### DIFF
--- a/lib/srfi/128/comparators.scm
+++ b/lib/srfi/128/comparators.scm
@@ -1,11 +1,23 @@
 
 (define-record-type Comparator
-  (make-comparator type-test equality ordering hash)
+  (%make-comparator% type-test equality ordering hash)
   comparator?
   (type-test comparator-type-test-predicate)
   (equality comparator-equality-predicate)
   (ordering comparator-ordering-predicate)
   (hash comparator-hash-function))
+
+(define (make-comparator type-test equality ordering hash)
+  (%make-comparator%
+    type-test
+    equality
+    ordering
+    (if (or (opcode? hash)
+            (not (procedure? hash))
+            (procedure-variadic? hash)
+            (> (procedure-arity hash) 1))
+        hash
+        (lambda (x . o) (hash x)))))
 
 (define-syntax hash-bound
   (er-macro-transformer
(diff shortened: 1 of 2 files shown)


lib/srfi/128/comparators.scm
```diff
@@ -1,12 +1,26 @@
...
+  (%make-comparator%
+    type-test
+    equality
+    ordering
+    (if (or (opcode? hash)
+            (not (procedure? hash))
+            (procedure-variadic? hash)
+            (not (= (procedure-arity hash) 1)))
+        hash
+        (case-lambda
```
ashinn: `case-lambda` is rather inefficient, and in this particular case `(lambda (x . o) (hash x))` is clearer anyway.

lib/srfi/128/comparators.scm
```diff
@@ -1,12 +1,26 @@
...
 
+(define (make-comparator type-test equality ordering hash)
+  (%make-comparator%
+    type-test
+    equality
+    ordering
+    (if (or (opcode? hash)
+            (not (procedure? hash))
+            (procedure-variadic? hash)
+            (not (= (procedure-arity hash) 1)))
```
ashinn: How about just `(> (procedure-arity hash) 1)`?

lib/srfi/128/comparators.scm
ar-nelson: Done.

lib/srfi/128/comparators.scm
ar-nelson: I figured this made more sense because the specific case it was fixing was a 1-argument function, but it doesn't really matter (0-argument functions will fail no matter what). Done.

ashinn:
Thanks!

PR #619, opened 2020-03-20
Implement SRFI 193: Command lines

lassik:
RnRS gives us a `command-line` procedure to get command line arguments. Since that procedure's specification is vague, I drafted a SRFI that specifies it in detail and adds some new related procedures. I'll submit the draft to the SRFI process once it's cleaned up.

Anyway, here's a Chibi implementation. It's simple, but needs to save the full `argv` from `run_main` into a Scheme variable so a few C additions are required. Would you like to include this kind of functionality in Chibi, and is the current implementation strategy reasonable?

Here's some sample output from the included `command-lines-test` script:

```scheme
(os-executable-file) => #f
(os-command-line) => ("chibi-scheme" "./command-lines-test" "foo" "bar")
(script-file) => "/Users/lassi/tmp/scheme/chibi/command-lines-test"
(script-directory) => "/Users/lassi/tmp/scheme/chibi/"
(command-name) => "command-lines-test"
(command-args) => ("foo" "bar")
(command-line) => ("command-lines-test" "foo" "bar")
(command-line-offset) => 1
```

### DIFF
--- a/main.c
+++ b/main.c
@@ -9,7 +9,8 @@
 #include "chibi/eval.h"
 #include "chibi/gc_heap.h"
 
-#define sexp_argv_symbol "command-line"
+#define sexp_command_line_symbol "command-line"
+#define sexp_raw_script_file_symbol "raw-script-file"
 
 #define sexp_import_prefix "(import ("
 #define sexp_import_suffix "))"
@@ -282,6 +283,14 @@ static void do_init_context (sexp* ctx, sexp* env, sexp_uint_t heap_size,
 #define init_context() if (! ctx) do {                                  \
       do_init_context(&ctx, &env, heap_size, heap_max_size, fold_case); \
       sexp_gc_preserve4(ctx, tmp, sym, args, env);                      \
+      sexp_set_parameter(                                               \
+          ctx, sexp_meta_env(ctx),                                      \
+          sym=sexp_intern(ctx, sexp_command_line_symbol, -1),           \
+          args=sexp_cons(ctx, sexp_c_string(ctx,"",-1), SEXP_NULL));    \
+      sexp_env_define(                                                  \
+          ctx, sexp_meta_env(ctx),                                      \
+          sym=sexp_intern(ctx, sexp_raw_script_file_symbol, -1),        \
+          SEXP_FALSE);                                                  \
     } while (0)
 
 #define load_init(bootp) if (! init_loaded++) do {                      \
@@ -306,7 +315,7 @@ sexp run_main (int argc, char **argv) {
   sexp_uint_t heap_size=0, heap_max_size=SEXP_MAXIMUM_HEAP_SIZE;
   sexp out=SEXP_FALSE, ctx=NULL, ls;
   sexp_gc_var4(tmp, sym, args, env);
-  args = SEXP_NULL;
+  args = NULL;
   env = NULL;
 
   /* SRFI 22: invoke `main` procedure by default if the interpreter is */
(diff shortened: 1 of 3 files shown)


lassik:
Output from a REPL session instead of a script:

```scheme
$ chibi-scheme
> (import (srfi command-lines))
WARNING: importing already defined binding: command-line
> (os-executable-file)
#f
> (os-command-line)
("chibi-scheme")
> (script-file)
#f
> (script-directory)
#f
> (command-name)
#f
> (command-args)
()
> (command-line)
("")
> (command-line-offset)
#f
```

As it says, the binding of `command-line` in the SRFI shadows the one in `(scheme process-context)`. The SRFI's `command-line` is compatible with the RnRS definition, so the same procedure can be used to implement both interfaces.

lassik:
`os-executable-file` requires a non-portable implementation for each OS. I have about 100 lines of C to do it for most of the OSes supported by Chibi. `#f` is an acceptable fallback.

lassik:
@ashinn Sorry, this PR fell through the cracks. Are you interested in bundling SRFI 193 with Chibi? If so, I'll finish it.

Note that slight changes to the C code are required - essentially the `main.c` changes that are in the current PR, perhaps modified to be more idiomatic.

main.c
```diff
@@ -306,7 +315,7 @@ sexp run_main (int argc, char **argv) {
   sexp_uint_t heap_size=0, heap_max_size=SEXP_MAXIMUM_HEAP_SIZE;
   sexp out=SEXP_FALSE, ctx=NULL, ls;
   sexp_gc_var4(tmp, sym, args, env);
-  args = SEXP_NULL;
+  args = NULL;
```
ashinn, line 318: Why are you changing this?  You set it to SEXP_NULL below anyway.

main.c
```diff
@@ -545,15 +554,22 @@ sexp run_main (int argc, char **argv) {
...
-    load_init(i < argc || main_symbol != NULL);
-    sexp_set_parameter(ctx, sexp_meta_env(ctx), sym=sexp_intern(ctx, sexp_argv_symbol, -1), args);
+      sexp_set_parameter(
+        ctx, sexp_meta_env(ctx),
+        sym=sexp_intern(ctx, sexp_command_line_symbol, -1),
+        args);
+      sexp_env_define(
+        ctx, sexp_meta_env(ctx),
+        sym=sexp_intern(ctx, sexp_raw_script_file_symbol, -1),
+        sexp_c_string(ctx,argv[i],-1));
```
ashinn, line 570: Why are repeating these settings here and above?

ashinn:
I think it's ok to bundle this since it can't be implemented portably.

main.c
lassik: Must be my mistake.

main.c
lassik: In `init_context` the two parameters are set to default values. Here they are set based on `argv[]`. If you can point at one central place where we could set them, that would be best.

ashinn:
This breaks `make test-run`.

lassik:
> This breaks `make test-run`.

Right, sorry about that. Adding a debug print to `tests/run/lib/hello.sld` shows that the args are `("Schemer!")` when the test code expects `(<program-name> "Schemer!")`. The command line is passed as an argument to the main procedure.

lassik:
I'm trying to fix the code but it's quite tricky. Is there a way to rewrite `main()` to call `init_context()` in one place only? Then we could set the `command-line` and `raw-script-file` parameters in the same place as well.

ashinn:
That code needs refactoring, but it may take me a while to get around to it.

I don't understand why you made so many changes though.
Just add the extra var where the current (command-line) is already set.

lassik:
Continued in PR #728

PR #622, opened 2020-03-25
(chibi net): add address-info-canonname

katterjohn:
`ai/canonname` has already been available in `(chibi net)` to request the canonical name of a host, but there wasn't a way to actually access it.  This PR exposes the `struct addrinfo` field `ai_canonname` via the getter `address-info-canonname`.

I chose the name `address-info-canonname` instead of something like `address-info-canonical-name` to match the name of `ai/canonname`.  I think their names should be consistent since they go together.


ashinn:
lgtm

PR #625, opened 2020-04-01
make-promise: return the argument if it's already a promise

katterjohn:
Chibi has not been conforming to R7RS here: `make-promise` needs to simply return its argument if it's already a promise instead of wrapping it again.

Chibi:
```
> (define foo (make-promise 1))
> (define bar (make-promise foo))
> (eqv? foo bar)
#f
> (force bar)
((#f . #<procedure #f>) promise)
> (force (force bar))
1
```

Chicken:
```
#;1> (define foo (make-promise 1))
#;2> (define bar (make-promise foo))
#;3> (eqv? foo bar)
#t
#;4> (force bar)
1
```

Kawa:
```
#|kawa:1|# (define foo (make-promise 1))
#|kawa:2|# (define bar (make-promise foo))
#|kawa:3|# (eqv? foo bar)
#t
#|kawa:4|# (force bar)
1
```

### DIFF
--- a/lib/scheme/lazy.sld
+++ b/lib/scheme/lazy.sld
@@ -4,4 +4,6 @@
   (export delay force delay-force make-promise promise?)
   (begin
     (define (make-promise x)
-      (delay x))))
+      (if (promise? x)
+          x
+          (delay x)))))
--- a/tests/r7rs-tests.scm
+++ b/tests/r7rs-tests.scm
@@ -325,6 +325,9 @@
     (let ((x (make-promise (+ 2 2))))
       (force x)
       (promise? x)))
+(test #t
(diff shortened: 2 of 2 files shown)


katterjohn:
There are currently a couple of AppVeyor build failures, but they appear to be intermittent and unrelated to the changes made in this PR.

katterjohn:
And here is the relevant portion of the description of `make-promise` from section 4.2.5 of R7RS-small:

> (make-promise _obj_)
>
> [...]  If _obj_ is already a promise, it is returned.

ashinn:
Sorry, wasn't paying attention and thought this was a bug rather than PR ^_^;
The fix is the same either way.

Issue #626, opened 2020-04-23
Help flag for the command line


lassik:
Would it be possible to add an official help flag that:

* prints the usage message to stdout
* prints nothing to stderr
* exits with code zero (success)

Currently `chibi-scheme -h` and `chibi-scheme --help` almost fit the bill by accident. `-h` is the heap size option and requires an argument, so if no argument is given it prints the usage (but exits with code 70 for error). `--help` is an unknown option and hence also prints usage and exits with code 70.

Is `-h` commonly used, and if not, would it be possible to change it to be the help flag (with another flag such as capital `-H` for the heap size)? Alternatively, could `--help` be officially recognized so that it exits with code zero?

ashinn:
How about `-?`?

lassik:
That would probably work but I'm not aware of any other program that has it. The de facto standard help flags are `-h`, `-help` (one dash) and `--help` (two dashes).

`?` is a glob metacharacter. I don't know whether any shells treat it as a special character in that position on the command line.

ashinn:
As long as you don't have a two letter file name beginning with a - (highly unlikely) the glob will fail to match and be passed as is.

I use -h.  I don't know how many others do.  I intentionally let this conflict with the standard help option, because as you note it does the right thing anyway.  The exit code doesn't seem to matter.

I guess I could support --help, though it would be the only option not using a single letter.

lassik:
It's somewhat common to add a special case for `--help` and `--version` as the only argument into getopt parsers.

lassik:
Thanks for implementing `--help` :)

Another distinction is that `--help` usually prints to stdout (since it's a valid invocation of the command), whereas on invalid command line flags the usage is written to stderr. This could be done by passing a `FILE*` argument of `stdout` or `stderr` to `sexp_usage()` depending on success/failure, and using `fprintf()` to write the usage to that stream.

PR #627, opened 2020-04-23
Write usage to stderr when bad options given

lassik:
Continuation of issue #626

### DIFF
--- a/main.c
+++ b/main.c
@@ -31,7 +31,8 @@
 
 #if SEXP_USE_MAIN_HELP
 void sexp_usage(int err) {
-  printf("usage: chibi-scheme [<options> ...] [<file> <args> ...]\n"
+  FILE *out = err ? stderr : stdout;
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -31,7 +31,8 @@
 
 #if SEXP_USE_MAIN_HELP
 void sexp_usage(int err) {
-  printf("usage: chibi-scheme [<options> ...] [<file> <args> ...]\n"
+  FILE *out = err ? stderr : stdout;
```
ashinn, line 34: can you just inline this?  chibi is written for very old compilers which don't allow var declarations mid-function.

lassik: I can, but isn't that declaration at the very beginning of the function?

lassik: I thought a separate declaration is clearer since there is some `#ifdef` magic in the middle of the `printf()` call, making it quite complex.

ashinn: Oh, sorry, nm.

PR #630, opened 2020-05-02
Implement SRFI 162.

mnieper:
From the abstract of SRFI 162:

>  This SRFI provides a few extra procedures and comparators to go with SRFI 128, Comparators. Implementers are urged to add them to their SRFI 128 libraries, for which reason they are not packaged as a separate library.

This pull request adds the sample implementation of SRFI 162 to Chibi-Scheme.

### DIFF
--- a/lib/srfi/125/test.sld
+++ b/lib/srfi/125/test.sld
@@ -5,15 +5,8 @@
           (chibi test))
   (begin
     (define (run-tests)
-      (define default-comparator (make-default-comparator))
       (define number-comparator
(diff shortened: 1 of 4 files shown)


ashinn:
Should we also provide `(srfi 162)`?

mnieper:
For feature detection it makes sense, but it's not part of SRFI 128/162,
which does not speak against it.

Alex Shinn <notifications@github.com> schrieb am So., 3. Mai 2020, 01:06:

PR #631, opened 2020-05-05
Implement SRFI 158.

mnieper:
This is more or less the sample implementation provided with SRFI 158. SRFI 158 is part of the tangerine edition of R7RS-large.


lassik:
Great to have this!

PR #635, opened 2020-05-11
make: fix uninstall target

Donaim:
uninstall target failed to remove
- share/srfi libraries
- man pages
- include headers

This patch repairs the makefile so that all chibi files
get removed correctly.

In case of man pages, they were actually installed
to a wrong place - "man1/man1" instead of "man1".


ashinn:
Thanks!

Issue #601, opened 2020-01-07
Crashes on the "triples" problem 3.69 solution from SICP.


lockywolf:
I have written the following code as a solution to SICP's problem 3.69:

```
(define-syntax cons-stream
  (syntax-rules ()
    ((cons-stream a b) (cons a (delay b)))))
(define (stream-car stream) (car stream))
(define (stream-cdr stream) (force (cdr stream)))
(define (stream-ref s n)
  (if (= n 0)
      (stream-car s)
      (stream-ref (stream-cdr s) (- n 1))))

(define (stream-map proc s)
  (if (stream-null? s)
      the-empty-stream
      (cons-stream (proc (stream-car s))
		   (stream-map proc (stream-cdr s)))))
(define stream-null? null?)
(define the-empty-stream '())
(define (stream-for-each proc s)
  (if (stream-null? s)
      'done
      (begin (proc (stream-car s))
	     (stream-for-each proc (stream-cdr s)))))
(define (display-stream s)
  (stream-for-each display-line s))

(define (display-line x)
  (newline)
  (display x))
(define (stream-enumerate-interval low high)
  (if (> low high)
      the-empty-stream
      (cons-stream
       low
       (stream-enumerate-interval (+ low 1) high))))
(define (stream-filter pred stream)
  (cond ((stream-null? stream) the-empty-stream)
	((pred (stream-car stream))
	 (cons-stream (stream-car stream)
		      (stream-filter pred
				     (stream-cdr stream))))
	(else (stream-filter pred (stream-cdr stream)))))

(define (stream-map proc . argstreams)
  (if (stream-null? (car argstreams))
      the-empty-stream
      (cons-stream
       (apply proc (map stream-car argstreams))
       (apply stream-map
	      (cons proc (map stream-cdr argstreams))))))

(define (add-streams s1 s2)
  (stream-map + s1 s2))
(define ones (cons-stream 1 ones))
(define zeros (cons-stream 0 zeros))
(define integers (cons-stream 1 (add-streams ones integers)))
(define (scale-stream stream factor)
  (stream-map (lambda (x) (* x factor)) stream))

(define (partial-sums s)
  (define ps (cons-stream (stream-car s) (add-streams ps (stream-cdr s))))
  ps)
(define (stream-print-n stream n)
    (do ((i 0 (+ i 1)))
      ((= i n) #t)
    (display (stream-ref stream i))
    (newline)))

(define (interleave s1 s2)
  (if (stream-null? s1)
[...]

PR #640, opened 2020-05-14
Prevent stack overflow in sexp_mark_one (issue #601)

Donaim:
Replace explicit recursion by heap allocations
in sexp_mark_one code.
This prevents crashes caused by stack overflow.
In particular, fixes issue #601.

As an optimization, allocate a fixed sized stack buffer first,
which should be enough for "normal" uses.
When that stack overflows, switch to heap.

Also, store "ranges" on the stack, instead of the actual sexp's,
using the fact that sexp's of a single parent are continous in memory.

This patch doesn't remove recursion on the context saves
because it didn't seem like they overflow in practice.
But changing that is simple having the stack interface.

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -225,7 +225,35 @@ int sexp_valid_object_p (sexp ctx, sexp x) {
 #define sexp_gc_pass_ctx(x)
 #endif
 
-void sexp_mark_one (sexp_gc_pass_ctx(sexp ctx) sexp* types, sexp x) {
+static void sexp_mark_stack_push (sexp ctx, sexp *start, sexp *end) {
+  struct sexp_mark_stack_ptr_t *stack = sexp_context_mark_stack(ctx);
+  struct sexp_mark_stack_ptr_t **ptr = &sexp_context_mark_stack_ptr(ctx);
+  struct sexp_mark_stack_ptr_t *old = *ptr;
+
+  if (old == NULL) {
+    *ptr = stack;
+  } else if (old >= stack && old + 1 < stack + SEXP_MARK_STACK_COUNT) {
+    (*ptr)++;
+  } else {
+    *ptr = malloc(sizeof(**ptr));
+  }
+
+  (*ptr)->start = start;
+  (*ptr)->end = end;
+  (*ptr)->prev = old;
+}
+
+static void sexp_mark_stack_pop (sexp ctx) {
+  struct sexp_mark_stack_ptr_t *stack = sexp_context_mark_stack(ctx);
+  struct sexp_mark_stack_ptr_t *old = sexp_context_mark_stack_ptr(ctx);
+
+  sexp_context_mark_stack_ptr(ctx) = old->prev;
+  if (!(old >= stack && old < stack + SEXP_MARK_STACK_COUNT)) {
+    free(old);
+  }
+}
+
+static void sexp_mark_one (sexp ctx, sexp* types, sexp x) {
   sexp_sint_t len;
   sexp t, *p, *q;
   struct sexp_gc_var_t *saves;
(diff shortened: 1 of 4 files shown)


Donaim:
I used a manual stack instead of max recursion depth argument because this solution does not rely on machine's stack size (and current position) and it is more extendable - ie. it could support runtime resizing that would be more efficient that using heap

gc.c
```diff
@@ -225,7 +225,46 @@ int sexp_valid_object_p (sexp ctx, sexp x) {
...
-void sexp_mark_one (sexp_gc_pass_ctx(sexp ctx) sexp* types, sexp x) {
+struct sexp_mark_stack_ptr_t {
+  sexp *start, *end;
+  struct sexp_mark_stack_ptr_t *next; /* TODO: remove for allocations on stack */
+};
+struct sexp_mark_stack_ptr_t *sexp_mark_stack_ptr = NULL;
+
+/* TODO: make resizable */
+#define SEXP_MARK_STACK_COUNT 1024
+struct sexp_mark_stack_ptr_t sexp_mark_stack[SEXP_MARK_STACK_COUNT];
```
ashinn, line 236: These variables should be a context global, or part of the heap.  Chibi supports multiple independent VMs running in parallel.

gc.c
```diff
@@ -225,7 +225,46 @@ int sexp_valid_object_p (sexp ctx, sexp x) {
...
 #endif
 
-void sexp_mark_one (sexp_gc_pass_ctx(sexp ctx) sexp* types, sexp x) {
+struct sexp_mark_stack_ptr_t {
+  sexp *start, *end;
+  struct sexp_mark_stack_ptr_t *next; /* TODO: remove for allocations on stack */
+};
+struct sexp_mark_stack_ptr_t *sexp_mark_stack_ptr = NULL;
+
+/* TODO: make resizable */
```
ashinn, line 234: This can go in features.h as a compile-time option, I don't think we need to worry about runtime.

gc.c
```diff
@@ -246,15 +285,30 @@ void sexp_mark_one (sexp_gc_pass_ctx(sexp ctx) sexp* types, sexp x) {
       q--;                      /* skip trailing immediates */
     while (p < q && *q == q[-1])
       q--;                      /* skip trailing duplicates */
-    while (p < q)
-      sexp_mark_one(sexp_gc_pass_ctx(ctx) types, *p++);
-    x = *p;
+    if (p < q) {
+      sexp_mark_stack_push(p, q);
```
ashinn, line 289: Thanks for this patch!  I appreciate the effort for a minimal diff.

One issue I see is that we mark all tails first before proceeding to the mark stack, making it highly likely the stack overflows.  If we want to keep the stack small we should be consuming it as we add to it.

gc.c
Donaim: Ok, will move the stack count there.

gc.c
Donaim: Ok, I will try to use context to store stack there

gc.c
Donaim: Thank you. Newer patch is not that minimal though:(

If I see it correctly, I don't think that we can use less memory here because we need to store "parent" pointers to be able to go back and recurse on parent nodes.
We free them as we go up, when done with children.
Storing parent pointers requires memory proportional to longest chain of sexps, which should be ~log(n) if they are balanced.

The problem here is tree traversal, where tree is sexp. Optimal space complexity for tree traversal is log(n), which is what we have.
This algorithm is actually pre-order depth first search (but we recurse on right-most node first and then go left-to-right...)
We could, teoretically, do traversal in O(1) if the tree was threaded, but I am sure it is not worth the effort to make sexp threaded.

Do you agree?

ashinn: Optimal space complexity on a _balanced_ tree is log(n).  This is a heap holding many heterogeneous data structures, such as linked lists which degenerate to linear.

We also don't need to store parent pointers.  You're representing the stack as a linked list, but we could also use an array and double the size as needed (or use a list of arrays).

There's also room for some small optimizations like checking for sexp_markedp _before_ pushing to the stack.  Let me play around with this.  I may merge and iterate later.

ashinn:
Thanks!  Tests pass, and no substantial diff in the benchmarks.

PR #643, opened 2020-05-18
Support for \u sequences in JSON

ekaitz-zarraga:
I'm not sure if this is the best way to do it but it works for simple \u sequences of UTF8 characters.

I'm not sure how should I make it compatible with compilation with SEXP_USE_UTF8_STRINGS set to false.

Ideas on that?


ashinn:
This looks fine.

Setting SEXP_USE_UTF8_STRINGS to false doesn't prevent encoding utf8 in strings, it just means basic string operations (string-ref, string-length, etc.) will work at the 8-bit byte level.  So the JSON will be read correctly (and written correctly), but possibly not processed correctly depending on what you do with it in Scheme.

PR #644, opened 2020-05-19
Support multiple \u sequences and add error checking

ekaitz-zarraga:
Some error checking when sequences have less than 4 numbers and some extra problems derived from `strtoll`'s cleverness.


lib/chibi/json.c
```diff
@@ -80,8 +81,12 @@ sexp parse_json_string (sexp ctx, sexp self, sexp str, const char* s, int* i, co
         from = to+1;
         break;
       case 'u':
-        strncpy(utf_tmp, s+to+1, 5);
-        tmp = sexp_make_string(ctx, sexp_make_fixnum(1), sexp_make_character(strtoll(utf_tmp, NULL, 16)));
+        strncpy(utf_tmp, s+to+1, 4);
+        tmp = sexp_make_string(ctx, sexp_make_fixnum(1), sexp_make_character(strtoll(utf_tmp, &utf_cursor, 16)));
```
ashinn, line 85: I guess strtol is sufficient?  This will never need more than 21 bits.

lib/chibi/json.c
```diff
@@ -80,8 +81,12 @@ sexp parse_json_string (sexp ctx, sexp self, sexp str, const char* s, int* i, co
         from = to+1;
         break;
       case 'u':
-        strncpy(utf_tmp, s+to+1, 5);
-        tmp = sexp_make_string(ctx, sexp_make_fixnum(1), sexp_make_character(strtoll(utf_tmp, NULL, 16)));
+        strncpy(utf_tmp, s+to+1, 4);
```
ashinn, line 84: Why the copy?  We can just use s.

lib/chibi/json.c
ekaitz-zarraga: I'm planning full utf support and in json that's done combining two \u sequences.

So maybe not now, but it can be useful later.

lib/chibi/json.c
```diff
@@ -80,8 +81,12 @@ sexp parse_json_string (sexp ctx, sexp self, sexp str, const char* s, int* i, co
...
+        tmp = sexp_make_string(ctx, sexp_make_fixnum(1), sexp_make_character(strtoll(utf_tmp, &utf_cursor, 16)));
+        if( utf_cursor-utf_tmp != 4 || utf_tmp[1] == 'x' || utf_tmp[1] == 'X' || utf_tmp[0] == '-'){
```
ashinn, line 86: I think you get unexpected results if the first character is 0 (octal).
Also, strtol will skip over leading whitespace.

nit: the style convention is `if (utf_cursor...`.

lib/chibi/json.c
ekaitz-zarraga: in the case of `\u12231243352462` strtol is going to take the whole number with no stop. We only need to parse the 4 chars and I didn't find a better way to do it.

lib/chibi/json.c
ekaitz-zarraga: I didn't think about leading space. I'll arrange something right now.

The octal... hmm not sure. I think if the base is set to 16 it should work fine.

ekaitz-zarraga:
I added a check based on `isxdigit` that is much more robust than what I did before and I removed the strtol check.

Hope this one looks better.

lib/chibi/json.c
ashinn: Full unicode is at most 21 bits, whatever the encoding.  That means just `strtol` is sufficient on any platform.

lib/chibi/json.c
```diff
@@ -80,10 +81,21 @@ sexp parse_json_string (sexp ctx, sexp self, sexp str, const char* s, int* i, co
         from = to+1;
         break;
       case 'u':
-        strncpy(utf_tmp, s+to+1, 5);
+        strncpy(utf_tmp, s+to+1, useq_len);
+        int err = 0;
+        for(int iter=0; iter!=useq_len; iter++){
+          if(!isxdigit(utf_tmp[iter])){
```
ashinn, line 87: nit: space after for and if.

ekaitz-zarraga:
Changed the order of the checks to avoid doing a string copy if input has errors, fixed if and for and moved to strol from strtoll

ekaitz-zarraga:
moved useq-len to an external define

ekaitz-zarraga:
I went supercrazy and I added support for coupled \u sequences for larger characters like here.

Sorry for the noise. I promise to stop adding stuff on top of this until I get some feedback.

utf8

Ty!

ashinn:
json-test.sld has a bunch of strings, can you put some escapes in there for coverage?

ekaitz-zarraga:
Some tests added.

ashinn:
There are minor outstanding comments - I'll just merge and fix myself.

ashinn:
Oh, sorry, the comments you hadn't addressed had been pending.  Ignore the submit just now, this was fixed in a subsequent commit.

PR #646, opened 2020-05-20
Fix indentation

ekaitz-zarraga:


lassik:
This finds all the inadvertent uses of tabs instead of spaces in Chibi source:

`git ls-files | grep -vi makefile | xargs file | grep ' text' | cut -d : -f 1 | xargs grep -n '\t'`

ashinn:
Thanks for catching this, I fixed it in another commit.

I may add some git hooks for basic formatting checks.

@lassik's suggestion is good too, though I would refine to:

`git ls-files | egrep -vi '(ma?ke?file|\.txt$)' | xargs file | grep ' text' | cut -d : -f 1 | xargs grep -n $'\t'`

Issue #590, opened 2019-11-07
Feature: JSON unparse


ekaitz-zarraga:
Hi,

As there's a JSON parse function I'm considering to develop a JSON unparse function.
Should I write it in C as json parser is or is it ok to go with Scheme + SRFI 159?
Any specific not about the design I should consider before spending time on it?

Thanks!

ashinn:
This is on my TODO list, and I was intending to write it in C, but PRs are welcome.

ekaitz-zarraga:
Cool, I'll try to implement it in Scheme this weekend with SRFI 159, and then consider rewriting it in C.

Thank you.

ekaitz-zarraga:
Hi @ashinn ,

I started to work on this on my own repo:
https://github.com/ekaitz-zarraga/chibi-scheme/pull/1

If you have some feedback please let me know. It would be very useful.
Once I'm more or less done I'll make a PR here.

ashinn:
This was merged (also I just updated this to work on port I/O).

PR #648, opened 2020-05-23
Unparse json feature

ekaitz-zarraga:
I made a JSON encoder functionality called `unparse-json` in order to solve #590 .
It's supposed to be the counterpart of `parse-json`, using the same types.

I'm not sure if the implementation is best in many cases. It's probably easy to improve in many points, that's why I separated everything in self-contained commits that are easier to review.

There are also some things I'm not sure about:

- JSON doesn't really limit the size of the numbers you can store on it, so I'm not sure if it's reasonable to add support for Bignums. Maybe we can convert them to float?
- I didn't add support for rationals in order to make the precision loss during the float conversion explicit and force the user to remind that JSON doesn't support them automatically.
- There are two implementations of the fixnum encoder, one uses alloca (bb1fdbb) and the other  (3745c16) doesn't but it's uglier in my opinion.
- Not sure about flonum implementation neither. You probably come up with something better.
- Portability is also a concern.

So it's a working solution but we'll probably need to make some changes.

I hope it's useful.

Thanks

### DIFF
--- a/lib/chibi/json.c
+++ b/lib/chibi/json.c
@@ -309,6 +309,11 @@ sexp sexp_parse_json (sexp ctx, sexp self, sexp_sint_t n, sexp str) {
 }
 
 
+
+
+sexp unparse_json (sexp ctx, sexp self, sexp obj);
+
+
 sexp unparse_json_string(sexp ctx, sexp self, const sexp obj){
   sexp_gc_var2(res, tmp);
   sexp_gc_preserve2(ctx, res, tmp);
@@ -357,6 +362,7 @@ sexp unparse_json_string(sexp ctx, sexp self, const sexp obj){
       chh = (0xD800 - (0x10000 >> 10) + ((ch) >> 10));
       chl = (0xDC00 + ((ch) & 0x3FF));
       if (chh > 0xFFFF || chl > 0xFFFF){
+        // TODO: Create custom unparse exception
         res = sexp_json_exception(ctx, self, "unable to encode character at", obj, i);
         return res;
       }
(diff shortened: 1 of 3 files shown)


ashinn:
We should probably change this to json-read and json-write in line with SRFI 180.

ekaitz-zarraga:
Well, srfi 180 doesn't work like this. It can handle files larger than memory (this lib can't) and it's designed to read from a port, so this different naming can help separate concepts.
If you are sure, I can arrange a new commit.

Did you have the chance to check the internals?

ashinn:
Actually, I specifically wrote this lib to process a data set that couldn't fit in memory.
However, I relied on the fact that the json was written one object per line.

It's worth extending to read directly from ports for convenience, although now I see
SRFI 180 allows generators (a terrible idea) and a tree fold (good idea, but still
difficult to support from C).

lib/chibi/json.c
```diff
@@ -307,10 +317,267 @@ sexp sexp_parse_json (sexp ctx, sexp self, sexp_sint_t n, sexp str) {
...
+  res = sexp_string_concatenate(ctx, res, SEXP_FALSE);
+except:
+  sexp_gc_release6(ctx);
+  return res;
+}
+
+sexp unparse_json (sexp ctx, sexp self, sexp obj) {
+  sexp_gc_var1(res);
+  sexp_gc_preserve1(ctx, res);
+  res = SEXP_NULL;
```
ashinn, line 541: The default return value should be SEXP_VOID.

ekaitz-zarraga: Done in 28f3641 
Thanks!

ashinn:
I'll probably switch to ports as a follow-up, but this is fine for now.

PR #649, opened 2020-05-24
Makefile.detect: Fix installation

okuoku:
`CHIBI_POSIX_COMPILED_LIBS` and `CHIBI_WIN32_COMPILED_LIBS` are intended to be added in `CHIBI_COMPILED_LIBS`, not `COMPILED_LIBS`.

It prevented install `(chibi time)` so/dll.

### DIFF
--- a/Makefile.detect
+++ b/Makefile.detect
@@ -179,9 +179,9 @@ CHIBI_POSIX_COMPILED_LIBS = lib/chibi/process$(SO) lib/chibi/time$(SO) \
 CHIBI_WIN32_COMPILED_LIBS = lib/chibi/win32/process-win32$(SO)
 
 ifndef EXCLUDE_POSIX_LIBS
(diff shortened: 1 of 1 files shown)


okuoku:
Found at: https://github.com/okuoku/yunibase/commit/62201c26c03b857935cf3d6d5473d78cbffbc34a

... How about drop Win32(except MSYS and Cygwin) from Makefile? Win32 is the only serious non-POSIX port that supported by the Makefile, and now we have CMake for it.

And, perhaps do we need to set up GitHub CI for POSIX platforms?

ashinn:
Thanks @okuoku!  Sorry about that, I wanted all GNU make extensions in just Makefile.detect.

If it works now I think there's no reason to drop Win32 from the Makefile.

okuoku:
Filed #654 to try to detect this.

PR #654, opened 2020-05-27
CI: Add Github Actions

okuoku:
This workflow tests against the standard Makefile build on Mac and Linux.

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -1,0 +1,23 @@
+name: CI
+
+on: [push, pull_request]
+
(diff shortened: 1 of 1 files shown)


okuoku:
I have tried Windows build as well, but it was waaay too slow compared to AppVeyor with unknown reason, FWIW it is even hosted by Microsoft itself..

`CI.yaml` contains Mac and Linux build rules and that will do:

1. `make`
2. `make test-all`
3. `sudo make install`
4. `cd tests && chibi-scheme r7rs-tests.scm`

PR #658, opened 2020-06-09
Use r7rs lazy primitives in srfi 41

lkstl:
Rewrite srfi 41 primitive layer in terms of r7rs primitives for delayed
evaluation.

### DIFF
--- a/lib/srfi/41.scm
+++ b/lib/srfi/41.scm
@@ -2,16 +2,16 @@
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; primitive layer adapted from SRFI 41 reference impl
 
-(define-record-type <stream>
+(define-record-type Stream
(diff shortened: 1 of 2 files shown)


lib/srfi/41.scm
```diff
@@ -1,73 +1,54 @@
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; primitive layer adapted from SRFI 41 reference impl
-;; TODO: rewrite this in terms or R7RS lazy primitives
 
-(define-record-type Stream
-  (make-stream box)
+(define-record-type <stream>
```
ashinn, line 5: Please revert this - Chibi consistently uses the convention of `Class-Name` instead of `<class-name>`.

The motivation is largely because this is used in both write `{Class-Name fields ...}` and match
`(@ Class-Name fields)`, so it's shorter and looks cleaner without the angle brackets.

ashinn:
Thanks!

PR #665, opened 2020-07-08
(chibi filesystem): add chown

katterjohn:
This PR adds `chown` to `(chibi filesystem)`.  `chmod` is already present.


okuoku:
Ugh, Windows do not have `chown`. Could you wrap it with `(cond-expand ((not windows) ...))` in `filesystem.stub`, just like `make-fifo` etc.

katterjohn:
OK, pushed.  Sorry about that!

PR #667, opened 2020-07-10
Fix some NaN comparisons: NaNs are not less than every fixnum

katterjohn:
R7RS section 6.2.6: regarding `=`, `<`, `<=`, `>`, `>=`: "If any of the arguments are +nan.0, all the predicates return #f."

However `(< +nan.0 n)` was yielding `#t` for fixnum `n`, and similarly for `(<= +nan.0 n)` and `(> n +nan.0)` and so on.  This also caused `(negative? +nan.0)` to return `#t`.

It just happened that NaNs were _less than_ all fixnums: if a conditional was written the other way around then NaNs would have been _greater than_ all fixnums instead.

I also looked at the flonum case: things like `(< +nan.0 0.0)`.  It seems like this case was sort of "accidentally" correct because of how comparisons are done, but if a conditional (similar to the one mentioned above) was written the other way around then NaNs would be both less than or equal to and greater than all or equal to all flonums (but still not equal).

So nothing was going wrong in the flonum case, but if left as-is it seems like this could cause some non-obvious problems later on since it seems somewhat delicate.  (Why delicate?  `sexp_compare` would return 1 when comparing a NaN to a flonum, but it happened to work because of how `sexp_apply` compares this return value to 0.)

For both the fixnum and flonum cases check for NaNs after getting the flonum values (just like in the bignum and ratio cases).

### DIFF
--- a/bignum.c
+++ b/bignum.c
@@ -1833,7 +1833,10 @@ sexp sexp_compare (sexp ctx, sexp a, sexp b) {
     case SEXP_NUM_FIX_FLO:
       f = sexp_fixnum_to_double(a);
       g = sexp_flonum_value(b);
-      r = sexp_make_fixnum(f < g ? -1 : f == g ? 0 : 1);
+      if (isnan(g))
+        r = sexp_xtype_exception(ctx, NULL, "can't compare NaN", b);
+      else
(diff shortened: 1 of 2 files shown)


katterjohn:
Strange.  For some reason everything seems to be staying queued in AppVeyor.

PR #688, opened 2020-08-24
(srfi 166 color) fix

krzygorz:
This fixes some issues with nested text attributes.
For example, running this snippet displays *both* abc and def as yellow:

```scheme
(show #t (on-blue (each (as-yellow "abc") "def")))
```
This one has a similar issue but additionally displays "def" as black:
```scheme
(show #t (as-red (each (as-underline (as-bold "abc") "def") "ghi")))
```

Basically, the problem was that the  `color` state variable kept information about the most recently changed attribute, but this was not always enough to recover the original style when ending a colored chunk of text.

I added a state variable for each independent text attribute (foreground color, background color, bold, italic, underlined). Procedures like `as-color`, `on-color` change the appropriate variable and then restore it.

Some tests had to be changed because when resetting an attribute to the default value, an escape code resetting only that specific value is emitted rather than the generic `\x1B;[0m`


ashinn:
Thanks!  LGTM.

PR #689, opened 2020-08-24
snow: accept "_" as version separator

pclouds:
$ tools/snow-chibi  implementations
    WARNING: Implementation gauche is an unsupported version,
    0.9.10_pre1, but at least 0.9.4 is required.
    cyclone
    sagittarius

The easy solution is also accept "_" as version separator and consider
"pre1" the forth component. This makes the warning go away, and I don't
think it'll affect version comparison on other schemes.


ashinn:
LGTM

PR #690, opened 2020-08-25
Initialize fileno count in sexp_make_fileno_op()

pclouds:
I'm hunting down some weird problem where (process->sexp) hangs. It is
because the pipe to the child's stdin is not closed properly, so the
child process never receives eof to terminate itself.

Sprinkling fprintf() further, it's found that the sexp_finalize_fileno()
inside sexp_finalize_port() is not called because the original fileno
count is zero, so after decrement it's negative one. Negative one is not
zero, so no closing.

This looks obviously bad. The fileno count sometimes is wrong right
after this line

    sexp_fileno_count(fileno)++;

in sexp_open_input_file_descriptor(). It's zero instead of one, meaning
it was negative one before. Bad bad bad. I don't think negative counts
are intentional.

From the look of it (and I'm very new to Chibi) it looks like we may
need to initialize fileno count in open-pipe function. And fileno sexp
is allocated in this function. This works for me (the bug is reliably
reproducible, so reliably verified) but I don't know if this is really
the right place...

### DIFF
--- a/sexp.c
+++ b/sexp.c
@@ -1865,6 +1865,7 @@ sexp sexp_make_fileno_op (sexp ctx, sexp self, sexp_sint_t n, sexp fd, sexp no_c
   if (sexp_filenop(res)) {
     sexp_fileno_no_closep(res) = sexp_truep(no_closep);
     sexp_fileno_openp(res) = 1;  /* not necessarily */
+    sexp_fileno_count(res) = 0;
     return res;
   }
 #endif
@@ -1873,6 +1874,7 @@ sexp sexp_make_fileno_op (sexp ctx, sexp self, sexp_sint_t n, sexp fd, sexp no_c
   if (!sexp_exceptionp(res)) {
     sexp_fileno_fd(res) = sexp_unbox_fixnum(fd);
     sexp_fileno_openp(res) = 1;
+    sexp_fileno_count(res) = 0;
     sexp_fileno_no_closep(res) = sexp_truep(no_closep);
 #if SEXP_USE_WEAK_REFERENCES
     sexp_insert_fileno(ctx, res);


sexp.c
```diff
@@ -1873,6 +1874,7 @@ sexp sexp_make_fileno_op (sexp ctx, sexp self, sexp_sint_t n, sexp fd, sexp no_c
   if (!sexp_exceptionp(res)) {
     sexp_fileno_fd(res) = sexp_unbox_fixnum(fd);
     sexp_fileno_openp(res) = 1;
+    sexp_fileno_count(res) = 0;
```
ashinn, line 1877: Strange, `sexp_try_alloc` should zero out the whole chunk, so this shouldn't be necessary.

sexp.c
```diff
@@ -1865,6 +1865,7 @@ sexp sexp_make_fileno_op (sexp ctx, sexp self, sexp_sint_t n, sexp fd, sexp no_c
   if (sexp_filenop(res)) {
     sexp_fileno_no_closep(res) = sexp_truep(no_closep);
     sexp_fileno_openp(res) = 1;  /* not necessarily */
+    sexp_fileno_count(res) = 0;
```
ashinn, line 1868: This can't be right.  What if it was 3 before?

ashinn:
The fileno_count is the count of (chibi-managed) ports backed by the fileno (directly, not counting indirectly via a FILE*).

So when creating a new fileno this count should definitely be initialized to 0 - I'll take a look why it might not be.
But if we find the fileno already, existed, we should leave the count unmodified.

It is then incremented/decremented by opening and closing ports on it.
Iff decremented to 0 do we finalize the fileno.

pclouds:
Any suggestions where I should look into? Because this always happens to me. I pushed a branch [1] if you want to try, but I'm not sure if it will happen for you. You'll need gauch, but I think you can substitute it with something that only exit when stdin is closed.

Since this is chibi-managed ports (made by open-pipe, then forked and close-output-port). That should be true. But somehow I get count zero before the decremen in sexp_finalize_port anyway.

[1] https://github.com/pclouds/chibi-scheme/tree/problem

pclouds:
OK more debugging, it looks like this only happens if sexp_make_input_port() is called with SEXP_USE_WEAK_REFERENCES defined. sexp_lookup_fileno() returns a valid fileno with count zero. Then the count is not incremented and sexp_finalize_port() where we unconditionally decrement the count to negative zero.

I don't know exactly what happens above those calls. But I guess we could simply check fileno_count zero before decrementing it. Not sure what we would do if it's already zero though...

ashinn:
Ah, sorry, my previous statement is incorrect.  We _do_ attempt to track the fileno for FILE* backed ports, but as you found are forgetting to inc the count in that case.

Let me know if the commit I pushed fixes this.

pclouds:
Oops, didn't see the last line. Yes it worked!

PR #691, opened 2020-08-27
snow: extract feature list for cond-expand

pclouds:
Currently a package's cond-expand contains the symbol of the target implementation and optionally more from config file. Execute a command (once) on target implementation to add their full feature list,making it available for each package to use.

All of these Schemes are tested. Larceny is just too annoying to get the feature list (no one-liner, and it could take a while) so Larceny stays the current behavior.

There is a small unrelated change here: the gosh command to get version. We don't need to call (exit), if stdin is closed properly (it should) then gosh should exit regardless.


pclouds:
First step at poking more into install targets that are not chibi. I'm looking at `library-installed?` next.

lib/chibi/snow/utils.scm
```diff
@@ -1,3 +1,71 @@
...
+      (unless cache
+        (set! cache (process->sexp cmd)))
+      cache)))
+
+(define (write-to-string x)
+  (call-with-output-string (lambda (out) (write x out))))
+
+(define known-implementations
+  `((chibi "chibi-scheme" (chibi-scheme -V) "0.7.3"
+           ,(make-memoized-process->sexp
```
ashinn, line 13: Why not just use `delay`?

lib/chibi/snow/utils.scm
```diff
@@ -1,3 +1,71 @@
...
+             '(chibi-scheme -e "(write (features))")))
```
ashinn, line 14: Or `-p'(features)'`.

lib/chibi/snow/utils.scm
pclouds: It didn't occur to me. I just didn't use delay/force often enough to remember.

PR #694, opened 2020-08-30
Small documentation fixes

lubgr:
These commits add the `(chibi optional)` documentation to the main page, where it was missing. The http link to SRFI 188 is fixed, and the documentation of the `trace` and `untrace` procedures of `(chibi trace)` is updated such that actual parameter names and the documented ones match. Let me know if there is room for improvement and/or whether I should squash these commits into a single one.


ashinn:
Thanks!

PR #702, opened 2020-09-04
snow: confirm before uploading packages

pclouds:
Uploading a package is an irreversible operation. It's not even about
accidentally leaking your secret sauce to the internet. You could upload
a package to snow-fort.org by accident and pollute the package name
space [1].

So let's ask the user first before going ahead uploading stuff. We only
ask once even if we're going to upload a dozen packages, so it's not
that annoying. The target repo is also shown in case you want to upload
to a custom repo and want to make sure it does so.

[1] I did (while attempting to uploading to a local snow-fort instance
    during testing). I guess `(chibi snow commands)` is forever mine
    now.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -947,9 +947,11 @@
    ((package-file? (car o))
     (if (not (every package-file? (cdr o)))
         (non-homogeneous))
-    (for-each
-     (lambda (package) (upload-package cfg spec package))
-     o))
+    (if (yes-or-no? cfg "Upload " o " to "
+                    (remote-uri cfg '(command package uri) "/?"))
+      (for-each
+       (lambda (package) (upload-package cfg spec package))
+       o)))
    (else
     (if (any package-file? (cdr o))
         (non-homogeneous))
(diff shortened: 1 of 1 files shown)


pclouds:
Two PSes

- I wonder if we should have an "un-upload" command. Even if the server side ends up rejecting it, it's still nice for local snow hosts.
- We probably should ask `[y/n/a/q]`. `a` for always yes for the rest, `q` for always no for the rest. I attempted it, but I think the cfg is immutable, so it's a bit tricky...

lib/chibi/snow/commands.scm
```diff
@@ -947,9 +947,11 @@
    ((package-file? (car o))
     (if (not (every package-file? (cdr o)))
         (non-homogeneous))
-    (for-each
-     (lambda (package) (upload-package cfg spec package))
-     o))
+    (if (yes-or-no? cfg "Upload " o " to "
+                    (remote-uri cfg '(command package uri) "/?"))
```
ashinn, line 951: Can you add a TODO: to include a summary (version, file size, etc.)?

ashinn:
I think un-upload is reasonable within some time frame (say 24-hours).

In the meantime people can ping me and I can manually remove packages.

pclouds:
> I think un-upload is reasonable within some time frame (say 24-hours).

OK I can try to add that.

> In the meantime people can ping me and I can manually remove packages.

Please remove `(chibi snow commands)` thank you :)

PR #709, opened 2020-10-14
Building docs depends on having the shared libraries available

gahr:
This unbreaks compiling with multiple make jobs.


ashinn:
Thanks!

PR #717, opened 2020-11-04
Fix bug in accumulating in (chibi loop)

bjoli:
Accumulating has a bug that makes only lists supported, due to it ignoring the init value and always using the empty list. This fixes that.


ashinn:
Thanks!

PR #721, opened 2020-11-30
Fix unaligned access in bytevector-{u,s}{16,32,64}-{ref,set!} 

ilammy:
Native code implementing bytevector accessors uses the following access pattern:

```c
*(intNN_t*)(base+offset)
```

This can result in so called "unaligned memory access" if the offset is not a multiple of 2, 4, 8, or if the base address has not been allocated at an aligned address (unlikely).

Most popular modern architectures—x86 and ARMs—allow unaligned memory accesses on the instruction level but they are typically performed a bit slower than properly aligned accesses.

On the other hand, there are architectures which do not allow unaligned memory accesses. Each load or store of a value longer than 1 byte should use properly aligned address on those architectures. That is, u16 should be loaded from even addresses, s32 should only be stored at an address which is a multiple of 4, and f64 (aka double) can be located only at addresses which are multiple of 8. If the address is not aligned, CPU raises an exception which typically results in the process being killed by the operating system with a SIGBUS signal.

SPARC is one of those architectures which are strict with alignment. The current access pattern in bytevector native code can result in unaligned accesses, which in turn results in crashes. This issue has been found in this way: Chibi test suite includes some tests for unaligned accesses and it failed on SPARC.

In order to avoid unaligned accesses, loads and stores need to be performed a bit differently, doing 'type punning' in a safe way, not just casting pointers which breaks strict aliasing rules.

The most portable and efficient way to do this is to use `memcpy()`. Compilers know about this trick and generate very efficient code here, avoiding the function call and using the most efficient instructions. (Obviously, only when optimizations are enabled.)

That is, given

```c
static inline uint32_t ref_u32(const void* p) {
  uint32_t v;
  memcpy(&v, p, sizeof(v));
  return v;
}
```

on x86 this will be compiled into a single `movl` instruction because x86 allows unaligned accesses, similar with ARM where this becomes a single `ldr` instruction. However, on RISC-V—another platform with strict alignment rules—this code compiles into 4 `lbu` instructions fetching 4 bytes, then some more arithmetic to stitch those bytes into a single 32-bit value.

Also, make sure that we don't miss anything and cover the rest of bytevector accessors with tests for unaligned memory access. Include both integers and floats, in little and big endian flavors.

### DIFF
--- a/lib/scheme/bytevector-test.sld
+++ b/lib/scheme/bytevector-test.sld
@@ -234,11 +234,18 @@
                (equal? (bytevector-s16-native-ref b 0)
                        (- 44444 65536)))))
 
-      (test-assert "bytevector-s16-{ref,set!} [unaligned]"
-        (let ((b (make-bytevector 3)))
+      (test-assert "bytevector-{u16,s16}-{ref,set!} [unaligned]"
+        (let ((b (make-bytevector 5)))
           (bytevector-s16-set! b 1 -77 (endianness little))
-          (equal? (bytevector-s16-ref b 1 (endianness little))
-                  -77)))
+          (bytevector-s16-set! b 3 -77 (endianness big))
+          (and (equal? (bytevector-s16-ref b 1 (endianness little))
+                       -77)
+               (equal? (bytevector-u16-ref b 1 (endianness little))
+                       (- 65536 77))
+               (equal? (bytevector-s16-ref b 3 (endianness big))
+                       -77)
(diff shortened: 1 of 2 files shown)


ilammy:
It would be nice if someone with actual SPARC hardware could test this (or run it on a SPARC VM which accurately emulates the alignment requirements).

ashinn:
Wow, thanks for this, taking a look.

PR #728, opened 2021-02-01
Re-implement SRFI 193 (Command line)

lassik:
Another attempt. This time the values are assigned in one place and `make test-run` passes.

### DIFF
--- a/main.c
+++ b/main.c
@@ -9,7 +9,8 @@
 #include "chibi/eval.h"
 #include "chibi/gc_heap.h"
 
-#define sexp_argv_symbol "command-line"
+#define sexp_command_line_symbol "command-line"
+#define sexp_raw_script_file_symbol "raw-script-file"
 
 #define sexp_import_prefix "(import ("
 #define sexp_import_suffix "))"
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -545,15 +546,22 @@ sexp run_main (int argc, char **argv) {
...
+      tmp = sexp_c_string(ctx,argv[i],-1);
+    sexp_env_define(
+      ctx, sexp_meta_env(ctx),
+      sym=sexp_intern(ctx, sexp_raw_script_file_symbol, -1), tmp);
+    for (j=argc-1; j>=i; j--)
+      args = sexp_cons(ctx, tmp=sexp_c_string(ctx,argv[j],-1), args);
+    if (main_symbol)
+      args = sexp_cons(ctx, tmp=sexp_c_string(ctx,main_symbol,-1), args);
+    if (args == SEXP_NULL)
+      args = sexp_cons(ctx, tmp=sexp_c_string(ctx,"",-1), args);
```
ashinn, line 561: What's this case for?  If there are no args, why `("")` instead of `()`?

lassik: RnRS requires there to be at least one string. Chez Scheme has a convention of using `("")` in the REPL case; given the RnRS stipulation, I think Chez's convention is the most logical and obvious one.

PR #734, opened 2021-03-23
Add a note about packaging.

lockywolf:
The most recent chibi didn't run even if installed, until I figured out that I need to supply both PREFIX and DESTDIR to the make install.

Is this information worth being added to the README?


ashinn:
Neither of these are required.  The README.md already explains the use of PREFIX.  DESTDIR is only needed by package managers, and both of these are widely used variables.

PR #737, opened 2021-04-01
string-split misses final empty split

arkoort:
`(string-split "123:" ":")` returns `("123")` instead of `("123" "")`


ashinn:
Thanks for the report, but this patch is incomplete - chibi wasn't properly handling the grammar argument at all.

Fixed separately.

arkoort:
Yes, it was quick fix of obviously incorrect behavior. Thank you for complete fix.

PR #740, opened 2021-04-15
Add SRFI 219: Define higher-order lambda

lassik:


lassik:
Do you want the tests as well?

ashinn:
Yes, this is pretty trivial but tests would be nice.

lib/srfi/219.sld
```diff
@@ -0,0 +1,11 @@
+(define-library (srfi 219)
+  (export define)
+  (import (rename (scheme base) (define define/native)))
+  (begin  (define-syntax define
+            (syntax-rules ()
+              ((define ((head . outer-args) . args) . body)
+               (define (head . outer-args) (lambda args . body)))
+              ((define (head . args) . body)
+               (define head (lambda args . body)))
+              ((define head . body)
```
ashinn, line 10: This last case should be `(define head value)`.

lassik:
OK, tests added.

I simplified the macro even further by removing the special case that expands `(define (name . args) . body)` into `(define name (lambda args . body))` since Chibi can do that by itself. That way, if you ever add special behavior into Chibi's main `(define (name ...))` expander, that behavior will be respected by SRFI 219 as well.

mnieper:
Wouldn't it make a lot of sense to make SRFI 219's "define" the default
export of "(scheme base)" for Schemes that do support it (like Chibi does
now)?

Am So., 18. Apr. 2021 um 11:58 Uhr schrieb Alex Shinn <
***@***.***>:

ashinn:
(scheme base) define doesn't support this curried syntax.

mnieper:
Indeed. So everyone who wants to use SRFI 219 does have to write something
like `(import (except (scheme base) define) (srfi 219))`.

As SRFI 219's `define` is backward-compatible with R7RS's define, I'd
therefore suggest moving SRFI 219's `define` into `(scheme base)`. While
this is more than what SRFI 219 requires, it makes using SRFI 219 so much
more comfortable that I wonder whether SRFI 219 will see much use without.

Am So., 18. Apr. 2021 um 14:38 Uhr schrieb Alex Shinn <
***@***.***>:

pclouds:
I would expect programs that use `(scheme xxx)` to work across all shemes that support them. Unless you suggest all r7rs schemes includes srfi-219 in `(scheme base)`, this is a bad move.

mnieper:
Am So., 18. Apr. 2021 um 15:09 Uhr schrieb Duy Nguyen <
***@***.***>:

> I would expect programs that use (scheme xxx) to work across all shemes
> that support them. Unless you suggest all r7rs schemes includes srfi-219 in (scheme
> base), this is a bad move.
>
I would suggest that all R7RS schemes that *ship* with SRFI 219 do this.
(To document what's needed, one can always write `(import (scheme base)
(srfi 219))`.

Of course, you make some points and it is the same point R6RS makes. But
that point (namely not only discouraging but explicitly forbidding
extensions to the base and standard libraries) is contrary to the
philosophy R7RS, which is actively opposing R6RS here.

lassik:
I tend to agree with @ashinn and @pclouds that it's better to require an explicit import. There may not be universal agreement that the SRFI 219 syntax is the right thing. Guile also ships it in a library that has to be imported explicitly. Only 7 Scheme implementations out of the 26 surveyed support it by default.

lassik:
The `(import (except (scheme base) define) (srfi 219))` suggested by @mnieper is the right thing to use in libraries.

For REPL use, maybe 219 could be included via `(import (chibi))`. That's a non-standard library so it doesn't matter if it includes non-standard syntax extensions. `(scheme base)` should not import the 219 version of `define` IMHO.

ashinn:
The point of minimal specification in the small language is not to support gratuitous incompatibility, but to allow for differences in implementation for strong reasons, such as genuinely innovative language design that can't be implemented as a library, or inability to support features due to unusual constraints or platform, etc.  This is worth explaining further, and I'll try to do so, but not in this issue comment,

The path to getting something changed in `(scheme base)` at this point is to get it into a SRFI, achieve widespread use of that  SRFI, and propose it for R8RS :)

PR #744, opened 2021-05-04
Fix emscripten build

pclouds:
It looks like the original -shared is ignored for some reason, so emcc
produces the final .wasm and .js code, the latter is named
chibi-scheme-static.bc, which cannot be linked to create
chibi.js. Moving -shared to STATICFLAGS seems to fix that.

I'm new to emscripten so don't trust me too much on this. But I tested
chibi.js and things still seem to work fine.


amirouche:
What is the version of emscripten you are using?

amirouche:
Six month ago it was working https://github.com/ashinn/chibi-scheme/commit/f13c826da0f305220a6f6446292d5b7f62be0997

Issue #751, opened 2021-06-23
(chibi math prime) Miller-Rabin primality test is not correctly implemented


wrog:
(in [/lib/chibi/math/prime.scm](/ashinn/chibi-scheme/blob/master/lib/chibi/math/prime.scm#L76))
`modular-root-of-one?` is not doing the right thing and this will result in further false positives (i.e., numbers deemed prime that actually aren't) beyond what the Miller-Rabin test is normally expected to produce.

First of all, the docstring:
```
;; Returns true iff any (modular-expt a odd*2^i n) for i=0..twos-1
;; returns 1 modulo n.
```
does not match up with what the routine is actually doing, which is that it returns true if **any** of the `a`^(`odd`*2^`i`) powers encountered are 1 or -1.

But what we actually need this to do is return true if either `a`^`odd` = 1 (and thence all squares of it are 1 and we're getting **no** information on square roots of 1) or that we encounter -1 first (and again all squares will be 1).  If we encounter 1 first and it's the square of something that's not -1 or 1, then 1 has a third square root, we're not in a field, and so we have a witness, `a`, to `n` not being prime.

An example where `modular-root-of-one?` is going wrong would be 
`n`=15 (`neg1`=14, `twos`=1, `odd`=7) and `a`=4, 
which is an actual witness to 15 not being prime because `a`^7=4 and `a`^14=1 (mod 15) and thus we have 4 being a third square root of 1 (along with 1 and 14)

Since this routine isn't exported and is only used once, it would probably best for the sake of readability to reverse the sense of it, rename it to `miller-rabin-composite-witness?`, i.e., so that it's returning true when it finds a witness.  Meaning we replace
```
(not (modular-root-of-one? ...))
```
with

```
(miller-rabin-composite-witness? ...)
```
in the body of `miller-rabin-composite?`  And then 

```
(define (miller-rabin-composite-witness? twos odd a n neg1)
  ;; Returns true iff (1) a^(n-1)!=1 mod n OR
  ;; (2) the powers of a include a square root of 1 other than {1, -1}.
  ;; (which then means Z/(n) cannot be a field and n cannot be prime)
  (let ((b (modular-expt a odd n)))
    (let lp ((i 0) (b b))
      (cond ((= b neg1)  #f)               ; -1 is an expected sqrt(1)
            ((= b 1)     (not (zero? i)))  ; !! previous b is bad sqrt(1)
            ((>= i twos))                  ; !! a^(n-1)!=1 mod n 
[...]

PR #752, opened 2021-06-28
(chibi math prime) fix miller-rabin-composite?, factor, etc (issue #751), add factor-alist

wrog:
This addresses all of issues with (chibi math prime) raised with [in #751](/ashinn/chibi-scheme/issues/751) thus far

- makes `miller-rabin-composite?` not miss viable witnesses
- `factor` is faster, `(factor 1) = ()`, `(factor 0)` raises an error
- `aliquot`, `totient` are way faster, also raise errors on n=0,  `(totient 1) = 1`, `(aliquot 1) = 0`

I've also included a new export `factor-alist`, that returns the factorization as an alist with multiplicities (rather than a list of repeated primes that you then have to count) that imho is more useful than `factor`.  (This is basically a feature request; it's in a separate commit, feel free to toss or rename.)

Also included is the 4-line version of `(factor ...)` that uses `make-factorizer` similar to the new `aliquot` and `totient`, but is commented out due to being ever-so-slightly slower but which you might prefer on aesthetic/smaller-code-size grounds.

### DIFF
--- a/lib/chibi/math/prime-test.sld
+++ b/lib/chibi/math/prime-test.sld
@@ -71,7 +71,7 @@
       (test #t (perfect? 496))
       (test #t (perfect? 8128))
 
-      (test '(1) (factor 1))
+      (test '() (factor 1))
       (test '(2) (factor 2))
       (test '(3) (factor 3))
       (test '(2 2) (factor 4))
@@ -86,6 +86,7 @@
       (test '(2 3 3) (factor 18))
       (test '(2 2 2 3 3) (factor 72))
       (test '(3 3 3 5 7) (factor 945))
+      (test-error (factor 0))
 
       (test 975 (aliquot 945))
 
--- a/lib/chibi/math/prime-test.sld
+++ b/lib/chibi/math/prime-test.sld
@@ -107,4 +107,7 @@
                         5772301760555853353
                         (* 2936546443 3213384203)))
 
+      (test "Miller-Rabin vs. Carmichael prime"
+            #t (miller-rabin-composite? 118901521))
+
       (test-end))))
--- a/lib/chibi/math/prime-test.sld
+++ b/lib/chibi/math/prime-test.sld
@@ -90,6 +90,11 @@
       (test '(3 3 3 5 7) (factor 945))
       (test-error (factor 0))
 
+      (test '() (factor-alist 1))
+      (test '((2 . 3) (3 . 2)) (factor-alist 72))
+      (test '((3 . 3) (5 . 1) (7 . 1)) (factor-alist 945))
+      (test-error (factor-alist 0))
+
       (test 0 (aliquot 1))
       (test 975 (aliquot 945))
       (test-error (aliquot 0))
--- a/lib/chibi/math/prime-test.sld
+++ b/lib/chibi/math/prime-test.sld
@@ -47,6 +47,7 @@
       (test 5 (prime-below 7))
       (test 797 (prime-below 808))
 
+      (test 1 (totient 1))
       (test 1 (totient 2))
       (test 2 (totient 3))
       (test 2 (totient 4))
@@ -56,6 +57,7 @@
       (test 4 (totient 8))
       (test 6 (totient 9))
       (test 4 (totient 10))
+      (test-error (totient 0))
 
       (test #f (perfect? 1))
       (test #f (perfect? 2))
@@ -88,7 +90,9 @@
       (test '(3 3 3 5 7) (factor 945))
       (test-error (factor 0))
 
+      (test 0 (aliquot 1))
       (test 975 (aliquot 945))
+      (test-error (aliquot 0))
 
       (do ((i 3 (+ i 2)))
           ((>= i 101))
--- a/lib/chibi/math/prime.scm
+++ b/lib/chibi/math/prime.scm
@@ -237,31 +237,10 @@
 
 ;;> Returns the factorization of \var{n} as a monotonically
 ;;> increasing list of primes.
-(define (factor n)
-  (when (zero? n)
-    (error "cannot factor 0"))
-  (factor-twos
-   n
-   (lambda (b n)
-     (let lp ((i 3) (ii 9) (n (abs n))
-              (res (append (make-list b 2)
-                           (if (negative? n) (list -1) '()))))
-       (cond
-        ((> ii n)
-         (reverse (if (= n 1) res (cons n res))))
-        ((zero? (remainder n i))
-         (lp i ii (quotient n i) (cons i res)))
-        (else
-         (lp (+ i 2)
-             (+ ii (* (+ i 1) 4))
-             n
-             res)))))))
-;; this version is slightly slower
-;;(define factor
-;;  (let ((rfactor (make-factorizer '()
-;;                  (lambda (l p k) (cons (make-list k p) l)))))
-;;    (lambda (n) (apply append! (reverse (rfactor n))))))
-
+(define factor
+  (let ((rfactor (make-factorizer '()
+                  (lambda (l p k) (cons (make-list k p) l)))))
+    (lambda (n) (concatenate! (reverse (rfactor n))))))
 
 ;;> The Euler totient φ(\var{n}) is the number of positive
 ;;> integers less than or equal to \var{n} that are
--- a/lib/chibi/math/prime.scm
+++ b/lib/chibi/math/prime.scm
@@ -226,6 +226,15 @@
                           (lp (+ i 2) (+ ii (* (+ i 1) 4))
                               n (put res i k)))))))))))))
 
+;;> Returns the factorization of \var{n} as a list of
+;;> elements of the form \scheme{(\var{p} . \var{k})},
+;;> where \var{p} is a prime factor
+;;> and \var{k} is its multiplicity.
+(define factor-alist
+  (let ((rfactor (make-factorizer '()
+                  (lambda (l p k) (cons (cons p k) l)))))
+    (lambda (n) (reverse (rfactor n)))))
+
 ;;> Returns the factorization of \var{n} as a monotonically
 ;;> increasing list of primes.
 (define (factor n)
--- a/lib/chibi/math/prime.scm
+++ b/lib/chibi/math/prime.scm
@@ -192,30 +192,24 @@
 ;;> Returns the factorization of \var{n} as a monotonically
 ;;> increasing list of primes.
 (define (factor n)
-  (cond
-   ((negative? n)
-    (cons -1 (factor (- n))))
-   ((<= n 2)
-    (list n))
-   (else
-    (let lp ((n n)
-             (res (list)))
-      (cond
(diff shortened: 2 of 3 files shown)


wrog:
Also noticed the documentation generator getting confused (e.g., `totient` appears multiple times in the generated doc).  Since this is already happening with `coprime?`, I'm guessing this is a lower priority thing with possibly other instances across other libraries, so maybe this should be a separate issue?

lib/chibi/math/prime.sld
```diff
@@ -1,11 +1,12 @@
 
 (define-library (chibi math prime)
-  (import (scheme base) (scheme inexact) (srfi 27))
+  (import (scheme base) (scheme inexact) (scheme case-lambda) (srfi 27))
```
ashinn, line 3: Prefer (chibi optional) which is both faster and more concise.

lib/chibi/math/prime.scm
```diff
@@ -175,57 +189,109 @@
...
+;; adds a prime factor \var{p} of multiplicity \var{k},
+;; returns a factorization function which returns the factorization
+;; of its non-zero integer argument \var{n} in this representation.
+;; The following optional arguments, if provided, specialize \var{put}
+;; for particular primes:
+;; \var{put2:}  for \var{p}=2,  called as \scheme{(\var{put2} \var{r} \var{k})})
+;; \var{put-1:} for \var{p}=-1, called as \scheme{(\var{put-1} \var{r})}).
+(define (make-factorizer r1 put . o)
+  (let-keywords o ((put2  (lambda (r k) (put r  2 k)))
+		   (put-1 (lambda (r)   (put r -1 1))))
```
ashinn, line 205: Keywords are overkill, you can just use let-optionals.  Actually, I should add define-opt* so you can

```
(define-opt* (make-factorizer r1 put (put2 (lambda (r k) (put r 2 k))) (put-1 (lambda (r) (put r -1 1))))
  ...)
```

And actually, since you only call this in 3 places and put2 and put-1 are different every time, you could also just make them required arguments and simplify things.

wrog: > put2 and put-1 are different every time
no, all of the `put`s are the same for `factor-alist` and the (commented out) short version of `factor` (i.e., apart from taking different arguments, which means having them both required would be painful).  

I used keywords because I wasn't sure what else `make-factorizer` might eventually be used for (was pondering if `for-each-factor` is the Right Thing).  Not wedded to them.

wrog:
Also unclear on why `modular-root-of-1?` needs to go on the export list (I don't see any use for it separate from the `miller-rabin-composite?`  At the very least, it's misnamed since, strictly speaking, since the multiplicative group is finite, *everything* is a root of 1; what's actually being checked here is that something has a particular expected *order* ((n-1)/2^k for some k) .

and it's not needed anymore now that `miller-rabin-witnesser` exists.

lib/chibi/math/prime.scm
wrog: ok, it's using `let-optionals` now.

> I should add define-opt*

I just looked at SRFI 89 and SRFI 177 and am sensing a huge rabbit hole here.

ashinn:
That export was an unintended commit while debugging - reverted.

I seem to recall the logic for `modular-root-of-one?` also checking -1 was just a speedup/simplification, since on the next loop iteration it would be 1 anyway.  So while the name is confusing, there is at worst an off-by-one error in the comment.

I will try to find a false positive.

lib/chibi/math/prime.scm
ashinn: Yes, that is indeed a rabbit hole.  It has some hopes of getting standardized soon though.

wrog: > I will try to find a false positive.

 and whump (see latest commit).   It turns out any [Carmichael number](https://mathworld.wolfram.com/CarmichaelNumber.html) with smallest prime factor ≥ 59 will do, anything of the form (6k+1)(12k+1)(18k+1) where all 3 of those factors are prime is a Carmichael number, and there are [scads of these](https://oeis.org/A046025).

wrog: It also occurs to me that sabotaging the random number generator is perhaps overkill.  

I.e., if the fact that  
 `(test #t (miller-rabin-composite? 118901521))` 
 fails 97% of the time on the old code even if we leave the random number generator alone is sufficient for your purposes (because the RNG only gets 4 tries to find a witness outside of 2..53 which means hitting something that's a multiple of one of those 3 prime factors (smallest is 241, so that's a 1/241 chance for that one; the others are harder),
 
 then we can reduce the test to 1 line and punt on the thread-safety issues inherent in bashing `random-integer` (i.e., if you care about ever trying to run tests concurrently on multiple cores or something ... buh)

lib/chibi/math/prime-test.sld
```diff
@@ -1,7 +1,22 @@
...
+      (let ((random-integer/actual random-integer))
+	(define (random-integer/unfortunate n)
+	  (let ((r (random-integer/actual n)))
+	    (if (avoid r)
+		(random-integer/unfortunate n)
+		r)))
+	;; TODO: this will not be thread safe
+	(dynamic-wind
+	    (lambda ()
+	      (set! random-integer random-integer/unfortunate))
```
ashinn, line 15: set!ing an imported variable is not legal R7RS.  I'm surprised I didn't make this an error.  It's ok for now, I'll fix the testing later.

lib/chibi/math/prime-test.sld
```diff
@@ -107,4 +128,11 @@
                         5772301760555853353
                         (* 2936546443 3213384203)))
 
+      (let ((n 118901521))
+	(with-random-stupidity
+	 (lambda (r) (not (coprime? r n)))
+	 (lambda ()
+	   (test "Miller-Rabine test on Carmichael prime"
+		 #t (miller-rabin-composite? n)))))
```
ashinn, line 136: Awesome, thanks!

lib/chibi/math/prime.scm
```diff
@@ -75,20 +76,43 @@
 
 (define (modular-root-of-one? twos odd a n neg1)
   ;; Returns true iff any (modular-expt a odd*2^i n) for i=0..twos-1
-  ;; returns 1 modulo n.
+  ;; returns 1 or -1 modulo n.
```
ashinn, line 79: It's fine to delete this function if not used anymore.

lib/chibi/math/prime.scm
```diff
@@ -175,57 +198,105 @@
...
+;; adds a prime factor \var{p} of multiplicity \var{k},
+;; returns a factorization function which returns the factorization
+;; of its non-zero integer argument \var{n} in this representation.
+;; The optional 3rd and 4th arguments, if provided, specialize \var{put}
+;; for particular primes:
+;; \var{put2}  for \var{p}=2,  called as \scheme{(\var{put2} \var{r} \var{k})})
+;; \var{put-1} for \var{p}=-1, called as \scheme{(\var{put-1} \var{r})}).
+(define (make-factorizer r1 put . o)
+  (let-optionals o ((put2  (lambda (r k) (put r  2 k)))
+		    (put-1 (lambda (r)   (put r -1 1))))
```
ashinn, line 214: The indentation is off here and in other places, it looks like you're using tabs.

Apologies if this isn't your preferred style, but the Chibi codebase forbids tabs.
You can get a warning about this if you run `make init-dev`.  Please use spaces.

lib/chibi/math/prime-test.sld
wrog: Yeah, I was in the middle of working out something complicated that respected hygiene and then I noticed `set!` just worked, which surprised me, too (Chez also forbids this).  I've been inactive since the R5RS era (... still kind of freaked out that they managed to agree on a module system).

And now I'm realizing that since my test case < 341550071728321 (where is this number from?) `random-integer` is not even being invoked at all, so I think I'll just remove this shit since it's evil.

lib/chibi/math/prime.scm
wrog: Actually, I hate tabs too.  It's just that too many of my projects were with people that preferred them so it became my default at some point.  Emacs makes this stuff invisible.  But yay, I can fix.

- is there a `CONTRIBUTING` file that I missed (or some other convention I don't know about?); might be worth adding one.

- How do you feel about rebasing?  (I'd just as soon redo things so that it's tabless from the beginning, and also the git log on this pull branch is getting pretty tangled)

lib/chibi/math/prime.scm
wrog: > we can reduce the test to that 1 line
doing this, especially now that I've figured out[ this is all deterministic anyway (see other thread)](https://github.com/ashinn/chibi-scheme/pull/752#discussion_r660947368).

lib/chibi/math/prime.scm
wrog: done

lib/chibi/math/prime.scm
```diff
@@ -175,57 +189,105 @@
...
+        (else
+         (lp (+ i 2)
+             (+ ii (* (+ i 1) 4))
+             n
+             res)))))))
+;; this version is slightly slower
+;;(define factor
+;;  (let ((rfactor (make-factorizer '()
+;;                  (lambda (l p k) (cons (make-list k p) l)))))
+;;    (lambda (n) (apply append! (reverse (rfactor n))))))
```
wrog, line 263: I think I should either 

1. kill these lines (if speed is all that matters) or 
2. swap this version in for the one above (if size matters more and a 5% slowdown is no big deal)

lib/chibi/math/prime-test.sld
ashinn: I can't recall where that number is from...

ashinn: Ah, it's from the observation that if n < 341,550,071,728,321, it is enough to test a = 2, 3, 5, 7, 11, 13, and 17 (http://www.math.dartmouth.edu/~carlp/PDF/paper25.pdf), although we're still doing extra work in those cases.

lib/chibi/math/prime.scm
ashinn: I'll add a note to the README.  Rebasing is fine.

lib/chibi/math/prime.scm
ashinn: At a 5% diff I'd prefer the smaller size here for Chibi.

lib/chibi/math/prime.scm
wrog: Just so that you know, github apparently [gives brownie points](https://github.com/ashinn/chibi-scheme/community) if there's [a separate CONTRIBUTING file](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors).

lib/chibi/math/prime.scm
wrog: ok,  done.

PR #764, opened 2021-07-25
Fix indentation of test runner output

jgesswein:
The output of the standard test runner does not look right as far as indentation and line wrapping is concerned (see [before.txt](https://github.com/ashinn/chibi-scheme/files/6873792/before.txt)). There is also some code that belongs to the standard test runner within the code that is not specific to a test runner. This PR fixes these issues (see [after.txt](https://github.com/ashinn/chibi-scheme/files/6873791/after.txt)) and allows to define test runners with output that is completely different from Chibi’s standard test runner. An example for a test runner is

```scheme
(define (report-test status info)
  (case status
    ((BEGIN) (display #\.))
    ((SKIP)  (display #\-))
    ((FAIL)  (display #\F))
    ((ERROR) (display #\E))))
(define report-group
  (case-lambda
   ((group) (if (= (test-group-ref group 'level) 0)
		(newline)))
   ((group parent) #f)))
(current-test-reporter report-test)
(current-test-group-reporter report-group)
```

This test runner creates a single line of text for a test run where “-” indicates a skipped test, “F” indicates a test failure and “E” an error.

```text
....-.F.E
```

### DIFF
--- a/lib/chibi/test.scm
+++ b/lib/chibi/test.scm
@@ -24,11 +24,11 @@
   (newline (current-error-port)))
 
 (define (exception-message exc)
-  (let* ((s (parameterize
-               ((current-output-port (open-output-string)))
-             (print-exception exc (current-output-port))
-             (get-output-string (current-output-port))))
+  (let* ((s (let ((p (open-output-string)))
+              (print-exception exc p)
+              (get-output-string p)))
          (n (- (string-length s) 1)))
+    ;; Strip the “ERROR: ” prefix if present
     (let loop ((i 0))
       (if (>= (+ i 2) n)
           (substring s 0 n)
@@ -185,6 +185,7 @@
 
 (define (test-run expect expr info)
   (let ((info (test-expand-info info)))
+    ((current-test-reporter) 'BEGIN info)
     (if (and (cond ((current-test-group)
                     => (lambda (g) (not (test-group-ref g 'skip-group?))))
                    (else #t))
@@ -256,43 +257,35 @@
 
 ;;> Begin testing a new group until the closing \scheme{(test-end)}.
 
-(define test-begin
-  (case-lambda
-   (()
-    (test-begin ""))
-   ((name)
-    (let* ((parent (current-test-group))
-           (group (make-test-group name parent)))
-      ((current-test-group-reporter) group parent)
-      (current-test-group group)))))
+(define-opt (test-begin (name ""))
+  (let* ((parent (current-test-group))
+         (group (make-test-group name parent)))
+    ((current-test-group-reporter) group parent)
+    (current-test-group group)))
 
 ;;> Ends testing group introduced with \scheme{(test-begin)}, and
 ;;> summarizes the results.  The \var{name} is optional, but if
 ;;> present should match the corresponding \scheme{test-begin} name,
 ;;> or a warning is printed.
 
-(define test-end
-  (case-lambda
-   (()
-    (test-end #f))
-   ((name)
-    (let ((group (current-test-group)))
-      (when group
-        (when (and name (not (equal? name (test-group-name group))))
-          (warning "mismatched test-end:" name (test-group-name group)))
-        ((current-test-group-reporter) group)
-        (let ((parent (test-group-ref group 'parent)))
-          (when parent
-            (test-group-inc! parent 'subgroups-count)
-            (cond
-             ((test-group-ref group 'skip-group?)
-              (test-group-inc! parent 'subgroups-skip))
-             ((and (zero? (test-group-ref group 'FAIL 0))
-                   (zero? (test-group-ref group 'ERROR 0))
-                   (= (test-group-ref group 'subgroups-pass 0)
-                      (test-group-ref group 'subgroups-count 0)))
-              (test-group-inc! parent 'subgroups-pass))))
-          (current-test-group parent)))))))
+(define-opt (test-end (name #f))
+  (let ((group (current-test-group)))
+    (when group
+      (when (and name (not (equal? name (test-group-name group))))
+        (warning "mismatched test-end:" name (test-group-name group)))
+      ((current-test-group-reporter) group)
+      (let ((parent (test-group-ref group 'parent)))
+        (when parent
+          (test-group-inc! parent 'subgroups-count)
+          (cond
+           ((test-group-ref group 'skip-group?)
+            (test-group-inc! parent 'subgroups-skip))
+           ((and (zero? (test-group-ref group 'FAIL 0))
+                 (zero? (test-group-ref group 'ERROR 0))
+                 (= (test-group-ref group 'subgroups-pass 0)
+                    (test-group-ref group 'subgroups-count 0)))
+            (test-group-inc! parent 'subgroups-pass))))
+        (current-test-group parent)))))
 
 ;;> Exits with a failure status if any tests have failed,
 ;;> and a successful status otherwise.
@@ -527,7 +520,6 @@
         info)))
 
 (define (test-default-applier expect expr info)
-  ((current-test-reporter) #f info)
   (let ((expect-val
          (guard
              (exn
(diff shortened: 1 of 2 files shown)


lib/chibi/test.scm
```diff
@@ -23,6 +23,20 @@
             args)
   (newline (current-error-port)))
 
+(define (exception-message exc)
+  (let* ((s (parameterize
+               ((current-output-port (open-output-string)))
+             (print-exception exc (current-output-port))
+             (get-output-string (current-output-port))))
```
ashinn, line 30: There's no need to parameterize the current-output-port, you can just create a local string port.

lib/chibi/test.scm
```diff
@@ -23,6 +23,20 @@
...
+         (n (- (string-length s) 1)))
+    (let loop ((i 0))
```
ashinn, line 32: Please add a comment: "Strip the ERROR: prefix if present."

lib/chibi/test.scm
```diff
@@ -242,66 +256,43 @@
...
-     (else
-      (display
-       (string-append
-        (make-string (or (test-group-indent-width group) 0)
-                     #\space)
-        (bold (string-append name ": "))))))
-    ;; set the current test group
-    (current-test-group group)))
+(define test-begin
+  (case-lambda
```
ashinn, line 260: Please don't use case-lambda, define-opt from (chibi optional) is faster and more concise.

lib/chibi/test.scm
```diff
@@ -242,66 +256,43 @@
...
-                 (test-group-inc! parent 'subgroups-skip))
-                ((and (zero? (test-group-ref group 'FAIL 0))
-                      (zero? (test-group-ref group 'ERROR 0))
-                      (= (test-group-ref group 'subgroups-pass 0)
-                         (test-group-ref group 'subgroups-count 0)))
-                 (test-group-inc! parent 'subgroups-pass))))
-             (current-test-group parent)
-             group))))))
+(define test-end
+  (case-lambda
```
ashinn, line 275: Please revert this case-lambda change as well.

lib/chibi/test.scm
```diff
@@ -324,28 +315,30 @@
...
-     (and (or (and parent (test-group-ref parent 'skip-group?))
-              (any (lambda (f) (f group)) (current-test-group-removers))
-              (and (null? (current-test-group-removers))
-                   (pair? (current-test-group-filters))))
-          (not (any (lambda (f) (f group)) (current-test-group-filters)))))
-    group))
+;; (name (prop . value) ...)
+(define (make-test-group name parent)
+  (let* ((g (list name))
+         (! (lambda (k v) (test-group-set! g k v))))
```
ashinn, line 321: This is fine, but in the future if you want a faster review of your changes please don't mix unrelated refactorings in with other changes :)

lib/chibi/test.scm
```diff
@@ -797,21 +818,23 @@
...
       => (lambda (parent)
            (test-group-set! parent 'trailing #f)
            (test-group-inc! parent 'total-pass pass)
            (test-group-inc! parent 'total-fail fail)
-           (test-group-inc! parent 'total-error err))))))
+           (test-group-inc! parent 'total-error err)))
+     (else
+      (when (zero? (test-group-ref group 'level))
+        (newline))))))
+(define test-default-group-reporter
```
ashinn, line 834: Add a newline.

lib/chibi/test.scm
```diff
@@ -712,8 +737,12 @@
            => (lambda (group) (test-group-set! group 'trailing #t))))))
   (flush-output-port)
   status)
+(define (test-default-reporter status info)
```
ashinn, line 740: Add a newline.

lib/chibi/test.scm
```diff
@@ -521,39 +527,34 @@
...
-          (let ((status
-                 (if (and (not (assq-ref info 'expect-error))
-                          (if (assq-ref info 'assertion)
-                              res
-                              ((current-test-comparator) expect-val res)))
-                     'PASS
-                     'FAIL))
-                (info `((result . ,res) (expected . ,expect-val) ,@info)))
-            ((current-test-reporter) status info)))))))
+  ((current-test-reporter) #f info)
```
ashinn, line 530: It seems weird to pass #f here to indicate a completely different semantics (indicating the start of a test) for the test reporter.  Better to split this into a separate parameter if needed.

lib/chibi/test.scm
```diff
@@ -797,21 +818,23 @@
...
            (test-group-set! parent 'trailing #f)
            (test-group-inc! parent 'total-pass pass)
            (test-group-inc! parent 'total-fail fail)
-           (test-group-inc! parent 'total-error err))))))
+           (test-group-inc! parent 'total-error err)))
+     (else
+      (when (zero? (test-group-ref group 'level))
+        (newline))))))
+(define test-default-group-reporter
+  (case-lambda
```
ashinn, line 835: This shouldn't use case-lambda, and in fact if you are splitting the logic should use two separate parameters.

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.scm
jgesswein: I used the special symbol 'BEGIN to indicate that a test is about to begin. For simplicity this is passed also for skipped tests. I did not use a separate parameter to keep the interface stable.

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.scm
jgesswein: Sorry about the noise. I’ll try to avoid this the next time.

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.scm
```diff
@@ -490,23 +476,36 @@
...
     (display " ")
     (flush-output-port)))
 
 (define (test-group-indent-width group)
   (let ((level (max 0 (+ 1 (- (test-group-ref group 'level 0)
                               (test-first-indentation))))))
-    (* 4 (min level (test-max-indentation)))))
+    (* (current-group-indent) (min level (test-max-indentation)))))
+
+(define indent-string
```
ashinn, line 500: This doesn't seem right as a global.  Or at least we should have some way of resetting it.  I think it's ok for now but can you add a note to that effect?

lib/chibi/test.sld
```diff
@@ -16,8 +16,9 @@
    current-test-epsilon current-test-comparator
    current-test-filters current-test-removers
    current-test-group-filters current-test-group-removers
-   current-column-width)
+   current-column-width current-group-indent)
   (import (scheme base)
+          (scheme case-lambda)
```
ashinn, line 21: You want to remove this case-lambda and unconditionally import (chibi optional).  It has a portable implementation on snow-fort.

lib/chibi/test.scm
jgesswein: Implemented

lib/chibi/test.sld
jgesswein: implemented

ashinn:
Thanks!

ashinn:
Hmmm... actually this seems to have broken the indentation that was working before (on make test-libs).

PR #773, opened 2021-08-18
Eliminate uses of strncpy()

lassik:
Can be replaced with memcpy() or snprintf().

### DIFF
--- a/eval.c
+++ b/eval.c
@@ -2433,7 +2433,6 @@ sexp sexp_load_standard_ports (sexp ctx, sexp env, FILE* in, FILE* out,
 }
 
 sexp sexp_load_standard_env (sexp ctx, sexp e, sexp version) {
-  int len;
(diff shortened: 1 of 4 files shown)


ashinn:
Why?

lassik:
`strncpy()` doesn't guarantee null termination: https://man.openbsd.org/strncpy#EXAMPLES 

If you know the length of your strings, you might as well use `memcpy()` directly. If you don't know, or don't want to check, `snprintf()` is safer, since it takes the buffer size and guarantees null termination.

`strlcpy()` is also safe, but it's non-standard.

ashinn:
There's enough space allocated, the 0 byte will be copied.  I'd allow a patch to switch everything to memcpy with a manual 0-byte setting, but I don't see the point of snprintf here (you're not checking the return value anyway).

PR #779, opened 2021-09-18
CMake linux fix: add linker flags -lm -lutil and -ldl (when configured for dynamic loading)

lubgr:
I tested the CMake build on Archlinux x86, where I discovered that linking fails due to the missing `-lm -lutil -ldl`  flags. Not sure how this slipped through - this PR fixes it (also tested on an ARM Archlinux system).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,9 @@ target_include_directories(libchibi-common
 target_link_libraries(libchibi-common INTERFACE
     ${BOEHMGC}
     $<$<CONFIG:SANITIZER>:-fsanitize=address,undefined>
-    $<$<PLATFORM_ID:Windows>:ws2_32>)
+    $<$<PLATFORM_ID:Windows>:ws2_32>
+    $<$<AND:$<PLATFORM_ID:Linux>,$<BOOL:${BUILD_SHARED_LIBS}>>:${CMAKE_DL_LIBS}>
+    $<$<PLATFORM_ID:Linux>:m>)
 
 #
 # Sources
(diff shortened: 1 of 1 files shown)


okuoku:
Hmm, is `libutil` should only required with pty functions thus it should go to `add_compiled_library` -- need to some way to pass dependency to it.

Currently,

```cmake
add_stubs_library(lib/chibi/pty.stub)
```

how about adding `LINK_LIBRARIES` parameter. `cmake_parse_argument` can parse such argument.

```cmake
add_stubs_library(lib/chibi/pty.stub LINK_LIBRARIES util)
```

Actually, .stub files includes metadata that indicates which library should be linked;

```scheme
(c-link "util")
```

but I'm not sure how to extract this.

lubgr:
That's a good suggestion! I think extracting the `(c-link "util")` reliably is non-trivial, while the `LINK_LIBRARIES` option is more doable. Needs some tweaks for the static build to work, as the dependencies needs to be collected for a collective linking when the `clib.c` gets generated and added to to `libchibi-scheme`, but it works.

ashinn:
I can add a flag for chibi-ffi to output the linker deps, but you need a working build first.

lubgr:
> I can add a flag for chibi-ffi to output the linker deps, but you need a working build first.

Thanks for offering your help on this! Given that there currenlty is only one `(c-link ...)` form, might it also be fine to keep this in manual sync for now?

ashinn:
Yes, it's fine to special case -lutil for now.

okuoku (approved):
> I can add a flag for chibi-ffi to output the linker deps, but you need a working build first.

When building shared library version, CMake build will build `chibi-scheme-bootstrap` first to run `chibi-ffi` thus that should be fine.

Difficult part is there is no way to describe `.stub` files themselves as "configure" dependency -- when `.stub` changed to link against another (native) library, we have to regenerate CMake build files. Unfortunately there's no straightforward way to do this.

Anyway, now the patch looks fine.

PR #780, opened 2021-09-29
Add missing constants to SRFI-144

jpellegrini:
The following constants were missing from Chibi's SRFI-144
implementation:

```
 fl-1/e
 fl-e-pi/4
 fl-1/log-2
 fl-log-3
 fl-log-pi
 fl-1/log-10
 fl-2pi
 fl-pi-squared
 fl-degree
 fl-gamma-1/2
 fl-gamma-1/3
 fl-gamma-2/3
```


jpellegrini:
Looks like in MinGW and MSVC we can't use `gamma` (although `log` and `sqrt` do work).
If you want I may change the PR to use the numerical values of the `gamma-...` constants

ashinn:
Shouldn't it be tgamma or lgamma_r?

jpellegrini:
Oh the SRFI requires true gamma,not log-gamma. I'll fix it

ashinn:
Thanks!

PR #791, opened 2021-11-30
Implement SRFI 229: Tagged Procedures

mnieper:
This patch implements SRFI 229 (Tagged Procedures) by adding an extra tag field to each procedure (closure). A new primitive lambda form that is generative is added so that different evaluations of the same procedure don't have to share a tag even if the procedure does have no free variables.

By setting SEXP_USE_TAGGED_PROCEDURES to 0, support is disabled.


ashinn:
There are a number of issues with this - notably it increases the number of core forms and increases the memory used by all procedures.

I pushed an alternative approach with no changes to the core.

mnieper:
> I pushed an alternative approach with no changes to the core.

That's nice!

PS Can you test the following, please?

```scheme
(test-not (procedure/tag? +))
```

In the code I wrote, `procedure?` was not sufficient to distinguish between operators and true procedures.

ashinn:
Good catch, guarded against that checking `(not (opcode? f))`.

PR #805, opened 2021-12-30
Identifier macros

dpk:
Adds identifier macros and variable transformers, plus support for `identifier-syntax` in `(chibi syntax-case)`.

See the comments on the diff.

Some tests/demos in `tests/syntax-tests.scm` wouldn’t go amiss.


lib/init-7.scm
```diff
@@ -143,13 +143,21 @@
     (lambda (expr use-env mac-env)
       (f expr mac-env))))
 
-(define er-macro-transformer
+(define er-macro-transformer*
```
dpk, line 146: I decided that only `er-macro-transformer` should keep the old behaviour of complaining when a syntax keyword is used outside of operator position. There are hardly any non-`er` macros in the Chibi standard library anyway.

sexp.c
```diff
@@ -2190,6 +2190,8 @@ sexp sexp_write_one (sexp ctx, sexp obj, sexp out, sexp_sint_t bound) {
 #endif
     case SEXP_PROCEDURE:
       sexp_write_string(ctx, "#<procedure ", out);
+      if (sexp_procedure_variable_transformer_p(obj))
+        sexp_write_string(ctx, "(variable-transformer) ", out);
```
dpk, line 2194: Perhaps omit this, or change it to just `#<variable-transformer ...>`? I mainly used it for debugging `make-variable-transformer`.

eval.c
```diff
@@ -399,6 +399,18 @@ static sexp sexp_make_macro (sexp ctx, sexp p, sexp e) {
...
 
+sexp sexp_make_variable_transformer_op (sexp ctx, sexp self, sexp_sint_t n, sexp base_proc) {
+  sexp_assert_type(ctx, sexp_procedurep, SEXP_PROCEDURE, base_proc);
+  if (sexp_procedure_variable_transformer_p(base_proc))
+    return base_proc;
+  sexp vt_proc = sexp_alloc_type(ctx, procedure, SEXP_PROCEDURE);
+  sexp_procedure_flags(vt_proc) = (char) (sexp_uint_t) sexp_make_fixnum (sexp_unbox_fixnum(sexp_procedure_flags(base_proc)) | SEXP_PROC_VARIABLE_TRANSFORMER);
+  sexp_procedure_num_args(vt_proc) = sexp_procedure_num_args(base_proc);
+  sexp_procedure_code(vt_proc) =  sexp_procedure_code(base_proc);
+  sexp_procedure_vars(vt_proc) = sexp_procedure_vars(base_proc);
```
dpk, line 410: I’m not entirely happy about how this copies all the fields by name, one-by-one, but I couldn’t work out how to make it work with `memcpy` (getting the right data and the right size of it behind the `sexp base_proc`)

eval.c
```diff
@@ -783,31 +811,43 @@ static sexp analyze_var_ref (sexp ctx, sexp x, sexp *varenv) {
...
-    } else if (sexp_exceptionp(value)) {
-      res = value;
-    } else if (sexp_immutablep(sexp_ref_cell(ref))
-               || (varenv && sexp_immutablep(varenv))) {
-      res = sexp_compile_error(ctx, "immutable binding", sexp_cadr(x));
+    cell = sexp_env_cell(ctx, sexp_context_env(ctx), sexp_cadr(x), 0);
+    op = cell ? sexp_cdr(cell) : NULL;
+    if (op && sexp_macrop(op)) {
+      if (!sexp_procedure_variable_transformer_p(sexp_macro_proc(op))) {
+        res = sexp_compile_error(ctx, "can't mutate a syntax keyword", sexp_cadr(x));
```
dpk, line 826: This incidentally fixes a segfault when you try to `set!` a syntax keyword on current versions of Chibi

eval.c
```diff
@@ -783,31 +811,43 @@ static sexp analyze_var_ref (sexp ctx, sexp x, sexp *varenv) {
...
+      } else {
+        res = analyze_macro_once(ctx, x, op, depth);
+        if (!sexp_exceptionp(res))
+          res = analyze(ctx, res, depth, defok);
```
dpk, line 830: This means that if a `set!` with a variable-transformer expands into another `set!`, it will add as many stack frames as there are `set!`s, unlike regular macro expansions which can use `goto loop` in the main `analyze` procedure to continue the expansion iteratively. The alternative would be to move at least the checking for the correct form, plus the check for a syntax keyword, into the appropriate clause of `analyze` itself.

mnieper:
With variable transformers being a new object, one can now reattack the problem of coexistence between `syntax-case` and Chibi's native transformers without any "ugly hacks" as Alex called them.

The idea would be to have `er-macro-transformer` and `sc-macro-transformer` return not a procedure object but a wrapped procedure object (as much as variable transformers are wrapped procedure objects). Chibi's native expander could then call the procedure object with as many arguments as are expected depending on the wrapper. Unwrapped objects would be called with a single argument for `syntax-case` compatibility.

ashinn:
Thanks Daphne!  I'll take a look.  I'm not necessarily opposed to identifier syntax in Chibi as I was in the small language standard, but would want some convincing use cases, and would be very picky about keeping the implementation as minimal as possible.

dpk:
> With variable transformers being a new object, one can now reattack the problem of coexistence between `syntax-case` and Chibi's native transformers without any "ugly hacks" as Alex called them.
> 
> The idea would be to have `er-macro-transformer` and `sc-macro-transformer` return not a procedure object but a wrapped procedure object (as much as variable transformers are wrapped procedure objects). Chibi's native expander could then call the procedure object with as many arguments as are expected depending on the wrapper. Unwrapped objects would be called with a single argument for `syntax-case` compatibility.

The current implementation, where variable-transformer-ness is just a one-bit flag on procedures, does not easily enable this.

mnieper:
> > With variable transformers being a new object, one can now reattack the problem of coexistence between `syntax-case` and Chibi's native transformers without any "ugly hacks" as Alex called them.
> > The idea would be to have `er-macro-transformer` and `sc-macro-transformer` return not a procedure object but a wrapped procedure object (as much as variable transformers are wrapped procedure objects). Chibi's native expander could then call the procedure object with as many arguments as are expected depending on the wrapper. Unwrapped objects would be called with a single argument for `syntax-case` compatibility.
> 
> The current implementation, where variable-transformer-ness is just a one-bit flag on procedures, does not easily enable this.

Can't one use the same machinery? ER/SC/RSC would set a similar bit on the procedure they return. A syntax-case transformer would be one where the ER/SC/RSC bit is cleared. The variable transformer bit is orthogonal.

dpk:
> Can't one use the same machinery? ER/SC/RSC would set a similar bit on the procedure they return. A syntax-case transformer would be one where the ER/SC/RSC bit is cleared. The variable transformer bit is orthogonal.

One could, yes. But with two boolean flags, one has four possible states for three valid states: regular procedure; syntax transformer not a variable transformer; variable transformer. Also, there are only 7 bits of flags and this would be the fourth such flag.  (That’s not a problem per se, but worth bearing in mind.)

Anyway, it could be done like that, I just wanted to point out the current patch doesn’t directly provide a means to do it.

mnieper:
> One could, yes. But with two boolean flags, one has four possible states for three valid states: regular procedure; syntax transformer not a variable transformer; variable transformer. Also, there are only 7 bits of flags and this would be the fourth such flag. (That’s not a problem per se, but worth bearing in mind.)

There are four valid states, aren't there? Variable transformer: yes/no. Expect three arguments instead of one as a procedure transformer: yes/no.

dpk:
> There are four valid states, aren't there? Variable transformer: yes/no. Expect three arguments instead of one as a procedure transformer: yes/no.

Oh, right. I had assumed you would only want to use syntax-case for identifier macros with this proposal, but there’s no reason for such a restriction.

mnieper:
> Thanks Daphne! I'll take a look. I'm not necessarily opposed to identifier syntax in Chibi as I was in the small language standard, but would want some convincing use cases, and would be very picky about keeping the implementation as minimal as possible.

The following list comprises a number of reasonable use cases, I think. It should be said in advance that identifier macros are in the same sense not necessary that `set!` or `call/cc` are not necessary; all of them can be eliminated through a global program transformation.

1. Procedure integration to enable more compiler optimizations like beta reduction and constant folding: For example, `map` could be defined as a macro, expanding into a (named) let when used in the form `(map proc list)` and expanding into a reference to a proper procedure when used in the form `map`. See `define-integrable` here: https://cisco.github.io/ChezScheme/csug9.5/syntax.html#./syntax:h1.
2. SRFIs like SRFI 190 become portably implementable when syntax parameters and identifier macros are available.
3. Fluids as in https://github.com/pre-srfi/fluids/blob/main/tests.scm become portably implementable in terms of ordinary parameters when identifier macros and identifier properties (SRFI 213) are available.
4. Keyword arguments with no runtime overhead: A "procedure" taking keyword arguments would be a macro keyword in disguise; only when the "procedure" is referenced directly, it evaluates to a first-class procedure; see 1.
5. A SRFI 17-like facility but without runtime overhead and without its controversial semantics. An example of its use could be:
```scheme
(let ((v (vector 1 2 3))
       (p (cons 'a 'b)))
  (let-syntax ((x (vector-at v 0))
                    (y (cdr-at p)))
    (set! x 4)
    (list x y v))) ; => (4 b #(4 2 3))
```
6. "Volatile" variables as in C when existing code expects variables so that volatiles cannot be emulated through procedure calls.
7. Constant variables realized with `identifier-syntax` so that a syntax error is raised when it is being `set!`.
8. Some solution to the `reference-barrier` problem of https://github.com/ashinn/chibi-scheme/pull/796#:
```scheme
(define-volatile !key)

(define (ephemeron-ref ephemeron !key)
  (let ((k (ephemeron-key ephemeron))
        (d (ephemeron-datum ephemeron)))
    (and (not (ephemeron-broken? ephemeron))
         (eq? !key k)
         d)))
```
Here, `define-volatile` adds an identifier property (à la SRFI 213) to `!key` meaning that every reference to `!key` mustn't be reordered (so it is effectively followed by `reference-barrier`). The `lambda` form would check for such marked identifiers and would realize them through syntax parameters bound to an identifier macro.

eval.c
mnieper: My two cents: If the macro expansion process so far happens in bounded C stack space, I think it is worth keeping this property.

dpk: Yes, I agree. Fortunately, a way to do it just occurred to me that neither involves moving half of `analyze_set` into `analyze`, nor using stack space for recursive transformations of variable transformer `set!` operations. I’ll try and implement it tomorrow.

dpk: Done f63348a

dpk:
Two comments on @mnieper’s contributions:

1. A separate flag for what arity a transformer procedures takes isn’t needed, as transformers are just procedures themselves, and arity is already stored for procedures. If procedure arity is 1, it gets a syntax object; if 3, it gets the form and two envs for syntactic closures.
2. It currently isn’t possible to bind an `identifier-syntax` macro to a syntax parameter because, with `(chibi syntax-case)` now ‘decoupled’ from the core, `syntax-parameterize` doesn’t know it needs to call `make-transformer` to make a transformer with the correct argument type. The same applies to `splicing-let-syntax` etc. It would need `(chibi syntax-case)` to re-export wrapped versions of every syntax binding form in the whole Chibi stdlib to fix this.

And furthermore to point 2,

3. Should `identifier-syntax` (appropriately rewritten in terms of `er-macro-transformer*`) be in `(chibi)` instead of `(chibi syntax-case)`? It doesn’t actually have anything to do with `syntax-case`’s regime of syntax objects and so on. It’s template-based like `syntax-rules`, and is really for identifier macros what `syntax-rules` is for regular S-expression macros. R6RS specifies it in the main spec document alongside `syntax-rules`, not in the library chapter about `syntax-case`. And without it, there’s no real convenient high-level way to take advantage of identifier macros, since `syntax-rules` itself can’t be used to write them.

eval.c
ashinn: You don't need this - just use make-procedure.

mnieper:
> Two comments on @mnieper’s contributions:
> 
>     1. A separate flag for what arity a transformer procedures takes isn’t needed, as transformers are just procedures themselves, and arity is already stored for procedures. If procedure arity is 1, it gets a syntax object; if 3, it gets the form and two envs for syntactic closures.

Unfortunately, it can't work what you have in mind because procedure arity is, in general, not well-defined.

Any transformer procedure could be written in the form `(lambda args ...)`, hiding the actual arity.

>     2. It currently isn’t possible to bind an `identifier-syntax` macro to a syntax parameter because, with `(chibi syntax-case)` now ‘decoupled’ from the core, `syntax-parameterize` doesn’t know it needs to call `make-transformer` to make a transformer with the correct argument type. The same applies to `splicing-let-syntax` etc. It would need `(chibi syntax-case)` to re-export wrapped versions of every syntax binding form in the whole Chibi stdlib to fix this.

With the - IMO unfortunate - relatively recent roll-back of the syntax-case integration, `(chibi syntax-case)` has lost a lot of its usefulness.

>     3. Should `identifier-syntax` (appropriately rewritten in terms of `er-macro-transformer*`) be in `(chibi)` instead of `(chibi syntax-case)`? It doesn’t actually have anything to do with `syntax-case`’s regime of syntax objects and so on. It’s template-based like `syntax-rules`, and is really for identifier macros what `syntax-rules` is for regular S-expression macros. R6RS specifies it in the main spec document alongside `syntax-rules`, not in the library chapter about `syntax-case`. And without it, there’s no real convenient high-level way to take advantage of identifier macros, since `syntax-rules` itself can’t be used to write them.

👍

dpk:
> Unfortunately, it can't work what you have in mind because procedure arity is, in general, not well-defined.
> 
> Any transformer procedure could be written in the form `(lambda args ...)`, hiding the actual arity.

I think in the specific case of macro transformer procedures it would be fine to throw an error at the Chibi core/C level if they don’t take exactly one or three arguments. `sc-macro-transformer` wraps the procedure it’s given into a wrapper that’s defined to take exactly three arguments, anyway, and I’m not sure there’s any practical use-case for defining a variadic `syntax-case` transformer procedure.

mnieper:
> > Unfortunately, it can't work what you have in mind because procedure arity is, in general, not well-defined.
> > Any transformer procedure could be written in the form `(lambda args ...)`, hiding the actual arity.
> 
> I think in the specific case of macro transformer procedures it would be fine to throw an error at the Chibi core/C level if they don’t take exactly one or three arguments. `sc-macro-transformer` wraps the procedure it’s given into a wrapper that’s defined to take exactly three arguments, anyway, and I’m not sure there’s any practical use-case for defining a variadic `syntax-case` transformer procedure.

Never write something that is incorrect in the general case if you can do better! While it seemingly makes no sense to make a `syntax-case` transformer procedure variadic, it is not too far-fetched to imagine a programmer writing a transformer as follows some time:

(define-syntax foo
  (lambda/doc (stx)
    "foo transformer procedure doc string"
    (syntax-case stx () ...)))

Where `lambda/doc` is imported from a hypothetical future `(srfi 300)`, whose implementation expands it along the following (very simplified) lines:

```
(case-lambda
  ((stx) (syntax-case stx () ...))
  (() "foo transformer procedure doc-string"))
```

(See also the discussion on the withdrawn SRFI 191 about arity inspection and why it is quite limited.)

eval.c
dpk: Oh, of course. Done eb8582f

dpk:
> 3. Should `identifier-syntax` (appropriately rewritten in terms of `er-macro-transformer*`) be in `(chibi)` instead of `(chibi syntax-case)`? It doesn’t actually have anything to do with `syntax-case`’s regime of syntax objects and so on. It’s template-based like `syntax-rules`, and is really for identifier macros what `syntax-rules` is for regular S-expression macros. R6RS specifies it in the main spec document alongside `syntax-rules`, not in the library chapter about `syntax-case`. And without it, there’s no real convenient high-level way to take advantage of identifier macros, since `syntax-rules` itself can’t be used to write them.

Okay, although I think it should be like this in the ideal case, it’s easier said than done because of template replacement, which needs to be shared with `syntax-rules`. It would be a moderately large change to the implementation of `syntax-rules` to restructure it to allow a second syntax definition form to use its template replacement system, and duplicating all that functionality would also be a bad idea, I think. Such changes to Chibi’s `syntax-rules` implementation are definitely out of the scope of this PR, so I’m punting on this for now.

However, I did realize that one could fix the problem with `syntax-case`-based `identifier-syntax` as a syntax parameter by just telling it to call `make-transformer` on the returned procedure itself: d769a79. This doesn’t fix the problem in the general case, and like Marc, I regret that the decoupling of `(chibi syntax-case)` from core makes `syntax-case` macros not usable in all contexts.

For one last use case that Marc didn’t mention (which now works with the latest change), one can for example exactly, but hygienically, replicate the classic *On Lisp*-style anaphoric macros with `identifier-syntax` and syntax parameters:

```scheme
(import (chibi syntax-case)
        (srfi 139))

(define-syntax-parameter it (identifier-syntax
                             (syntax-error "anaphor not bound")))

(define-syntax aif
  (syntax-rules ()
    ((_ test conseq) (aif test conseq (if #f #f)))
    ((_ test conseq alter)
     (let ((%it test))
       (syntax-parameterize ((it (identifier-syntax %it)))
         (if %it conseq alter))))))
```

Without identifier macros, you would have to awkwardly write `(it)` and not `it`. I think syntax parameters and identifier macros are a particularly nice combination in this regard.

I’m marking this ready for merge, though I would welcome further comments from @ashinn on the patch itself if improvements can still be made to the implementation.

mnieper:
> Okay, although I think it should be like this in the ideal case, it’s easier said than done because of template replacement, which needs to be shared with `syntax-rules`. It would be a moderately large change to the implementation of `syntax-rules` to restructure it to allow a second syntax definition form to use its template replacement system, and duplicating all that functionality would also be a bad idea, I think. Such changes to Chibi’s `syntax-rules` implementation are definitely out of the scope of this PR, so I’m punting on this for now.

That's a good point. Eventually, the syntax-case pattern matcher in `(chibi syntax-case)` (and the `syntax-rules` implementation derived from it) should be merged with the `syntax-rules` pattern matcher in `init-7.scm`, I think. But this would mean that some parts of the `syntax-case` implementation in `(chibi syntax-case)` would have to be moved into `init-7.scm`.

> For one last use case that Marc didn’t mention (which now works with the latest change), one can for example exactly, but hygienically, replicate the classic _On Lisp_-style anaphoric macros with `identifier-syntax` and syntax parameters:
> 
> ```scheme
> (import (chibi syntax-case)
>         (srfi 139))
> 
> (define-syntax-parameter it (identifier-syntax
>                              (syntax-error "anaphor not bound")))
> 
> (define-syntax aif
>   (syntax-rules ()
>     ((_ test conseq) (aif test conseq (if #f #f)))
>     ((_ test conseq alter)
>      (let ((%it test))
>        (syntax-parameterize ((it (identifier-syntax %it)))
>          (if %it conseq alter))))))
> ```
> 
> Without identifier macros, you would have to awkwardly write `(it)` and not `it`. I think syntax parameters and identifier macros are a particularly nice combination in this regard.

Indeed; identifier macros make syntax parameters more appealing. It should be noted, though, that it is not quite clear whether the `it` in the anaphoric `if` should be a syntax parameter. The following alternative implementation should be considered as well (for simplicity I am leaving out the one-armed case):

```scheme
(define-syntax aif
  (lambda (stx)
    (syntax-case stx ()
      ((k test expr1 expr2)
       (with-syntax ((it (datum->syntax #'k 'it)))
         #'(let ((it test))
              (if it expr1 expr2)))))))
```

This version is unhygienic (in a controlled sense) but composes better (think of `expr1` expanding into a macro use that itself uses the anaphoric `if`).

The disadvantage of the second version is that the binding of any identifier `it` in scope is being overwritten. This can, in turn, be corrected by making the name `it` parameterizable (using SRFI 213):

```scheme
(define-syntax anaphoric-key (syntax-rules ()))

(define-syntax with-anaphoric
  (lambda (stx)
    (syntax-case stx ()
      ((k it body1 ... body2)
       (identifier? #'it)
       #'(let* ()
           (define-property aif anaphoric-key 'it)
           (let* () body1 ... body2))))))

(define-syntax aif
  (lambda (stx)
    (lambda (lookup)
      (syntax-case stx ()
        ((k test expr1 expr2)
         (with-syntax ((it (datum->syntax #'k (lookup #'aif #'anaphoric-key)))
           #'(let ((it test))
                (if it expr1 expr2))))))))

(define-property aif anaphoric-key 'it)
```

Example code using it:

```scheme
(with-anaphoric es
  (aif 42 es #f)) ; => 42
```

lib/init-7.scm
johnwcowan: Is that really the Right Thing?  Allowing er-macro names to appear in operand position allows procedure/datum (e.g. keyword) macros on top of ER as well as syntax-case.  The e argument would be an identifier, which can't otherwise happen.

dpk: You seem to have understood this the wrong way around:  `er-macro-transformer` *will* complain if the `expr` argument is not a pair. This means all but one or two of the macros in the Chibi standard library will continue to give a helpful error message if their keywords appear outside operator position, also including ones written with `syntax-rules` (since `syntax-rules` expands into an `er-macro-transformer`). There’s a new `er-macro-transformer*` which doesn’t do this check explicitly and so can be used to write identifier macros. In addition, `sc-macro-transformer` and `rsc-macro-transformer` also don’t explicitly check their input types. In practice, this means the error message you’ll get for using those macros, if they weren’t intended to be used as identifier macros, will just be a somewhat confusing ‘can’t take car/cdr of a non-pair’ — this affects one or two macros in the Chibi `lib` directory but none in `(chibi)` or the R7RS libraries.

mnieper: I think Daphne made this decision because of the legacy code using `er-macro-transformer` in Chibi, which assumes a list argument for the form to be transformed. While it wouldn't be incorrect to extend the domain of `er-macro-transformer` silently, it would reduce the quality of error messages in Chibi.

Should `er-macro-transformer` be standardized one day, it would make sense to standardize the version that also handles identifier macros, of course.

An alternative to Daphne's approach would be to go through all ER macros in Chibi. That won't take as long as it sounds, I think.

dpk: Not only the code in Chibi, but the fact that `er-macro-transformer` is certainly the most commonly-used of the low-level macro transformation systems in code outside of Chibi itself.

dpk:
An additional use case: some apps may want to have Emacs Lisp-style `autoload` functionality for some variables and procedure names etc, which identifier macros make possible.

ashinn:
Thanks Daphne and Marc for the use cases.  I can imagine many more myself, but those are nice lists and should be preserved somewhere.

What I was interested in though was immediate applications by Chibi users in real-world programs.  I get the impression that this was not the motivation of the patch, and it is rather a vehicle of the large language standardization process.  Since there is an open vote on this exact issue I won't make any decision on accepting the patch until after the vote - I don't want the existence or not of support in Chibi to be part of the argument.

Suffice it to say, if you're able to tweak the macro expander itself identifier syntax is simple to implement.

mnieper:
Thanks for your response, Alex.

In some sense, this patch can nevertheless become part of an argument in the open vote on identifier macros in R7RS-large because people may be inclined more towards it if they know that they can be relatively easily implemented in existing Scheme systems that do not yet support them.

dpk:
Okay, I’ve merged `core-identifier-syntax` into this branch.

By the way, I noticed that `make test-all` doesn’t actually run `tests/syntax-tests.scm`. Is this deliberate?

ashinn:
OK, for the record I think the R6RS semantics of requiring make-variable-transformer only for the set! case but not the ref case is weird.  I understand _why_ they did it - the ref case will just error regardless if not handled, but the set! case could be misinterpreted as an operator call - but from the programmer's perspective it makes more sense to group them together.

Anyway, I'm fine with this change but it has to touch the core less.  I can implement it myself or you can update this PR with the following:

* move make-variable-transformer and identifier syntax out of the core, into (chibi ast)
* allow ref matching for all SC/ER macros, but if there's an error wrap it or sexp_warn that the macro was in a non-operator position

dpk:
> move make-variable-transformer and identifier syntax out of the core, into (chibi ast)

I can do this for `make-variable-transformer` (although it seems a bit weird to me that the flag exists and is recognized in core but the procedure to set it is in a library), but `identifier-syntax` needs access to one or another version of pattern/template replacement code. That means either the version in `(chibi)` for `syntax-rules` (as implemented in d55d6c6..920ba20) or the version in `(chibi syntax-case)` (as originally in 4a3c7ea and d769a79). `(chibi ast)` cannot depend on `(chibi syntax-case)`, because `(chibi syntax-case)` already depends on `(chibi ast)`. That leaves two options:

- Restore the `(chibi syntax-case)` version of `identifier-syntax` into `(chibi syntax-case)`
- Leave the changes to the `(chibi)` `syntax-rules` implementation in place that *allow* it to be used to implement `identifier-syntax` as well, but move the actual `identifier-syntax` implementation into `(chibi ast)`. This seems even weirder to me than having support for the variable transformer bit in core but not the op to set it.

> allow ref matching for all SC/ER macros, but if there's an error wrap it or sexp_warn that the macro was in a non-operator position

Okay, can do. (So in principle, just eliminate the distinction between `er-macro-transformer` and `er-macro-transformer*`?)

ashinn:
Yes, remove the `er-macro-transformer[*]` distinction.

There's nothing weird about moving code out of the core which isn't used by anything in the standard distribution.

identifier-syntax is a shorthand that can be defined portably in terms of syntax-rules.  Probably the current syntax-rules makes assumptions that expr is a list, but you can just guard against that.  If you're having trouble,  R6RS defines it in terms of syntax-case so you can just add that definition to (chibi syntax-case) and I'll add a TODO to make it work with syntax-rules.

dpk:
It can only with difficulty be implemented directly in terms of (unchanged) `syntax-rules`: as you say, `syntax-rules` requires and assumes that the expr argument is a list, as R7RS requires both patterns and input expressions to `syntax-rules` clauses macros to be lists. Furthermore, the first element in each pattern and in the input expression is ignored anyway (also according to R7RS, for reasons I don’t entirely understand, other than that it’s copied from R5RS which did not have `_`).

> The ⟨pattern⟩ in a ⟨syntax rule⟩ is a list ⟨pattern⟩ whose first element is an identifier. […] The keyword at the beginning of the pattern in a ⟨syntax rule⟩ is not involved in the matching and is considered neither a pattern variable nor a literal identifier. (§ 4.3.2)

Have you looked at my changes to the `syntax-rules` implementation? I know they’re not pretty, but they’re needed to work around the differences between (the two-clause form of) `identifier-syntax` and `syntax-rules`. `syntax-rules` is deliberately defined, even in R6RS, so as not to be useful for writing identifier macros. The bailiwicks of `identifier-syntax` and `syntax-rules` are thus mutually exclusive. And making the two ‘compatible’ that way would break the R7RS semantics, by making the keyword at the beginning of the pattern significant. (You also can’t solve this by deciding whether to ignore the initial keyword at macro expand time rather than at macro definition time — that was my initial approach, though I’ve forgotten the exact way it was broken.)

I’ll put `identifier-syntax` in the `syntax-case`-based implementation back in `(chibi syntax-case)` for now, and move the corresponding tests to the syntax-case test file.

ashinn:
Yep, it's an error for the pattern to be a non-list, which means we're free to do what we want in that case and can implement a syntax-rules which supports identifier-syntax.

Matching `(set! var expr)` would be a more grey area (we're not _supposed_ to match the `set!` but it would never match accidentally in practice) except that by wrapping in make-variable-transformer we're free to change the semantics.

dpk:
> Yep, it's an error for the pattern to be a non-list, which means we're free to do what we want in that case and can implement a syntax-rules which supports identifier-syntax.

Making this change to `syntax-rules` causes the `r7rs-tests.scm` test `count-to-2_` to fail:

```
    (list (count-to-2_ _ _) (count-to-2_) (count-to-2_ a b) (count-.…  [ FAIL]
        expected (2 0 fail fail) but got (fail fail fail fail)
        on line 514 of file "tests/r7rs-tests.scm"
```

The change that was made (and which fixes the failure if reverted):

```diff
diff --git a/lib/init-7.scm b/lib/init-7.scm
index 8a3ea91f..18992857 100644
--- a/lib/init-7.scm
+++ b/lib/init-7.scm
@@ -873,8 +873,8 @@
       (set! count (+ count 1))
       (rename (string->symbol (string-append s (%number->string count)))))
     (define (expand-pattern pat tmpl)
-      (let lp ((p (cdr pat))
-               (x (list _cdr _expr))
+      (let lp ((p pat)
+               (x _expr)
                (dim 0)
                (vars '())
                (k (lambda (vars)
```

ashinn:
I've pushed the proper fix for reference variables for syntax-rules.  set! support needs some more thought.

dpk:
I will leave you to work out how to solve that, then. I already went through this six weeks ago (including, as I recall, trying the exact solution you just proposed), and ended up with the solution in d55d6c6.

lib/init-7.scm
```diff
@@ -873,8 +873,8 @@
       (set! count (+ count 1))
       (rename (string->symbol (string-append s (%number->string count)))))
     (define (expand-pattern pat tmpl)
-      (let lp ((p (cdr pat))
-               (x (list _cdr _expr))
+      (let lp ((p pat)
+               (x _expr)
```
ashinn, line 877: Did you forget to revert this?

dpk: I haven’t changed anything since you pushed your proposed change.

PR #806, opened 2022-01-02
Expose construction of foreign procedures.

dpapavas:
This PR provides similar functionality to #803 and may be preferable to it.  It is simpler in terms of implementation (and working with procedures, as opposed to opcodes, seems to have some additional benefits, at least for my use case).  On the other hand, it looks like it might be less efficient, although my understanding of the VM is too slight to be able to make such judgements.  Even if that is the case, I would argue that the functionality would nevertheless be useful, for use cases where efficiency is not of primary importance.

Attached, you'll also find [a couple of test files](https://github.com/ashinn/chibi-scheme/files/7799607/test.tar.gz), in case they prove useful.

### DIFF
--- a/include/chibi/eval.h
+++ b/include/chibi/eval.h
@@ -129,6 +129,7 @@ SEXP_API sexp sexp_make_lit (sexp ctx, sexp value);
 SEXP_API sexp sexp_make_opcode (sexp, sexp, sexp, sexp, sexp, sexp, sexp, sexp, sexp, sexp, sexp, sexp, sexp_proc1);
 SEXP_API sexp sexp_make_procedure_op (sexp ctx, sexp self, sexp_sint_t n, sexp flags, sexp num_args, sexp bc, sexp vars);
 SEXP_API sexp sexp_define_foreign_aux (sexp ctx, sexp env, const char *name, int num_args, int flags, const char *fname, sexp_proc1 f, sexp data);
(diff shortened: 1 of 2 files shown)


ashinn:
This approach looks like a minimal change with no performance impact.  Could you make the tests proper test cases in `tests/ffi/ffi-tests.scm`?

dpapavas:
Sure, but `ffi-tests.scm` seems to contain only tests related to the C FFI and the newly introduced `sexp_define_foreign_proc`/`sexp_define_foreign_proc_rest` are not integrated into the FFI.  Perhaps adding a test in `tests/foreign/` would be more appropriate?

ashinn:
Right, but hold off I think we can make this simpler.

dpapavas:
Is there something I need to do about this?  I'm in no particular hurry to get this implemented, so I don't mean to rush it or anything; just wanted to make sure you're not waiting for something from me, that I've missed.

ashinn:
Sorry, been busy, the delay is on my end.  Will try to look at this over the weekend.

ashinn:
Note I may move the utility procedures elsewhere later, e.g. inlined by the FFI.

dpapavas:
Thanks for merging this!  When you say "move them elsewhere", would one still be able to call `sexp_define_foreign_proc[_rest]` in order to export C procedures manually, or would it be necessary to use the FFI (assuming this means using the `chibi-ffi` script?

PR #809, opened 2022-01-09
Build fails with SEXP_USE_MALLOC

dpapavas:
This is really an issue, not a PR.  It's not something I particularly need, but during the investigation of a bug I suspected to be due to the GC, I tried defining `SEXP_USE_MALLOC`, to see whether the behavior would change.  This resulted in a couple of C compile errors, which I fixed, but building just fails a bit later when building Scheme library stubs.  I didn't pursue the failure further, but I though I'd report it.

(Unless I'm missing something, this seems to be broken for quite a while - at least since 0.8, so if it has gone undetected until now, perhaps doing away with it altogether would be appropriate.)

### DIFF
--- a/gc.c
+++ b/gc.c
@@ -37,7 +37,7 @@ static sexp_heap sexp_heap_last (sexp_heap h) {
   return h;
 }
 
-#if !SEXP_USE_FIXED_CHUNK_SIZE_HEAPS
+#if !SEXP_USE_FIXED_CHUNK_SIZE_HEAPS && !SEXP_USE_MALLOC
 static size_t sexp_heap_total_size (sexp_heap h) {
   size_t total_size = 0;
   for (; h; h=h->next)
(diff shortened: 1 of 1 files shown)


ashinn:
This option is a hack which isn't suitable to any non-trivial program, so I don't guarantee support for it.

If you're having trouble with GC I recommend SEXP_USE_BOEHM or SEXP_USE_CONSERVATIVE_GC at a first pass.  There are other GC options to investigate further.

dpapavas:
Thanks for the tips.  I had tried to `SEXP_USE_CONSERVATIVE_GC` and it did indeed make the problem "go away", but I had already found out by then, that it was due to not preserving some values I should have been preserving (I suppose, because what needs to be preserved and what doesn't, isn't entirely clear).

I noticed you merged the PR.  Does the source build for you correctly with it?  As I mentioned, it only fixes the C side of the problem for me;  I still get Scheme errors when the FFI processes certain library stubs.

Issue #816, opened 2022-03-15
Remaining identifier macro issues


dpk:
- `(chibi syntax-case)` still tries to import `make-variable-transformer` from `(chibi)` rather than `(chibi ast)` (this is my fault for forgetting that module still depends on `make-variable-transformer` even though `identifier-syntax` is gone from it!)
- `identifier-syntax` is still broken since the change to `syntax-rules` in 4382b9d:

  ```scheme
  (import (chibi ast))
  (define-syntax foo (identifier-syntax (_ 'ref) ((set! _ _) 'set)))
  foo ;=> ERROR: cdr: not a pair: foo
  (foo) ;=> ERROR: non procedure application: ref (this is correct, but the next line isn’t)
  (foo x y) ;=> set
  ```

PR #818, opened 2022-03-16
Fix definition of full-match? (Proposed fix for #816)

dpk:
Full matching applies for all syntax rules if any of the syntax patterns is not a pair. Maintains the old behaviour for conformant R7RS `syntax-rules` macros (without breaking the `count-to-2_` test), but also allows identifier macros, and fully fixes `identifier-syntax`.


ashinn:
Thanks Daphne, this is simpler than what I had in mind.  I guess it won't work if they try to write a set! macro without ref support, but nobody would do that.  Let me take a look.

Issue #802, opened 2021-12-27
Support SRFI 206


dpk:
As far as I can tell, the groundwork for what would be needed to support SRFI 206 is already there in the `(auto)` magic module. But, despite much poking about in `meta-7.scm`, I can’t work out how to implement `define-auxiliary-syntax` itself.

`(define-auxiliary-syntax foo bar)` in definition context should be equivalent to an import like `(import (rename (only (auto) bar) (bar foo)))` I think. That works in `syntax-rules`, at least:

```scheme
> (import (only (auto) bar))
> (import (rename (only (auto) bar) (bar foo)))
> (define-syntax is-bar
    (syntax-rules (bar)
      ((_ bar) #t)
      ((_ _) #f)))
> (is-bar bar)
#t
> (is-bar foo)
#t
> (is-bar frotz)
#f
```

but according to `free-identifier=?` from `(chibi syntax-case)`, `#'foo` and `#'bar` are still distinct after doing this. That looks like a separate bug.

ashinn:
The magic happens in the module system itself, not through a primitive like `define-auxiliary-syntax`.  `(auto)` is treated just like a normal library, but in `resolve-import` we check for it specially and ensure any new bindings are defined (using `eval`).

So you'd have to rely on `import` being allowed in a definition context and use something like:

```
(define-syntax define-auxiliary-syntax
  (er-macro-transformer
   (lambda (expr rename compare)
     ;; intentionally not renaming only/auto keywords since currently
     ;; module names and keywords are not hygienic
     `(,(rename 'import)
       (rename (only (auto) ,(cadr expr))
               (,(cadr expr) ,(car (cddr expr))))))))
```

dpk:
Unfortunately, that can’t work in all cases because of internal definitions. Using this definition corrected to the actual syntax of SRFI 206:

```scheme
(define-library (srfi 206)
  (import (scheme base)
          (only (meta) repl-import)
          (only (chibi) er-macro-transformer))
  (export define-auxiliary-syntax)
  
  (begin
    (define-syntax define-auxiliary-syntax
      (er-macro-transformer
       (lambda (expr rename compare)
         (let* ((keyword (cadr expr))
                (symbol (if (pair? (cddr expr))
                            (car (cddr expr))
                            keyword)))
           ;; intentionally not renaming only/auto keywords since currently
           ;; module names and keywords are not hygienic
           `(,(rename 'repl-import)
             (rename (only (auto) ,symbol)
[...]

PR #820, opened 2022-04-03
Implement SRFI 213 (Identifier properties)

dpk:
This implementation is really not very good at all, but at the moment I can’t see a better way without significantly digging into Chibi’s core. It should suffice for most simple uses of identifier properties, but has some funky scoping and other undesirable behaviour.

Here are the problems:
- Identifier properties are stored in a global nested hash table. This means that any env containing a cell which has a property attached to its binding will never get GC’d. (Chibi still seems to lack weak hash tables which would fix this problem?)
- Because define-property isn’t recognized by the core macro expander, it sees it as an expression rather than as a definition.
- The implementation of `capture-lookup` is arguably not faithful to the requirement that ‘`capture-lookup` must be substitutable by the identity (lambda (proc) proc) when used in macro transformers, but I’m not really sure what that’s meant to mean. (Why have `capture-lookup` at all if it’s practically an identity function?)

Because of these and possibly for other reasons, it passes nearly none of Jim Rees’s advanced [tests](https://github.com/scheme-requests-for-implementation/srfi-213/blob/master/tests/srfi-213-tests.sch) it can with this definition of `identifier-property`:

```scheme
(define-syntax identifier-property
  (er-macro-transformer
   (lambda (expr rename compare)
     (capture-lookup
      (lambda (lookup)
        (list (rename 'quote)
              (lookup (cadr expr) (car (cddr expr)))))))))
```

- [x] ‘The second define-property overwrites the first one in a spliced context’ passes 
- [ ] ‘The macro picks up the outer definition’ fails for reasons I don’t understand
- [ ] ‘Demonstrate that properties can be fetched at higher phases too’ skipped because it involves mixing `with-syntax` and `syntax-rules`
- [ ] ‘property settings behave like macro definitions’ fails because of the second problem described above
- [x] ‘the 'key' identifier can be "introduced" by a macro and the property is still accessible by the calling code’
- [ ] ‘demonstrates that when the id is "introduced" by the macro the property is not accessible via the un-marked id’ fails for probably also because of the second problem described above too
- [ ] ‘Properties are attached to identifiers and not to bindings’ skipped because I don’t understand it (the spec document says they’re attached to bindings?) and Chibi doesn’t support internal aliases, see #802

The basic tests, above the ones by Jim Rees, (again suitably modified) should all pass, though.

But given how compromised this whole thing is, especially considering this is another change to the implementations of the core macro transformer constructors (to support `capture-lookup` as the spec intends), I’m not sure if merging would actually be a good idea.

### DIFF
--- a/lib/chibi/syntax-case.scm
+++ b/lib/chibi/syntax-case.scm
@@ -23,7 +23,7 @@
             (old-renamer (current-renamer)))
         (current-usage-environment use-env)
         (current-renamer (make-renamer mac-env))
-        (let ((result (transformer expr)))
+        (let ((result (capture-lookup-wrap use-env (transformer expr))))
           (current-usage-environment old-use-env)
           (current-renamer old-renamer)
           result))))
--- a/lib/init-7.scm
+++ b/lib/init-7.scm
@@ -133,22 +133,32 @@
        '()))
     rename))
 
+(define capture-lookup-wrap
+  (lambda (use-env expansion)
+    (if (procedure? expansion)
+        (expansion use-env)
+        expansion)))
+
 (define sc-macro-transformer
   (lambda (f)
     (lambda (expr use-env mac-env)
-      (close-syntax (f expr use-env) mac-env))))
+      (close-syntax
+       (capture-lookup-wrap use-env (f expr use-env))
+       mac-env))))
 
 (define rsc-macro-transformer
   (lambda (f)
     (lambda (expr use-env mac-env)
-      (f expr mac-env))))
+      (capture-lookup-wrap use-env (f expr mac-env)))))
 
 (define er-macro-transformer
   (lambda (f)
     (lambda (expr use-env mac-env)
-      (f expr
-         (make-renamer mac-env)
-         (lambda (x y) (identifier=? use-env x use-env y))))))
+      (capture-lookup-wrap
+       use-env
+       (f expr
+          (make-renamer mac-env)
+          (lambda (x y) (identifier=? use-env x use-env y)))))))
 
 (define-syntax cond
   (er-macro-transformer
--- a/lib/srfi/213.scm
+++ b/lib/srfi/213.scm
@@ -1,0 +1,50 @@
+(define id-comparator
+  (make-pair-comparator eq-comparator eq-comparator))
+(define *identifier-properties* (make-hash-table id-comparator))
+
+(define-syntax define-property
+  (lambda (expr use-env mac-env)
+    (let* ((id (cadr expr))
+           (id-cell
+            (or (env-cell use-env id)
+                (error "can't attach property to unbound identifier" id)))
+           (key (car (cddr expr)))
+           (key-cell
+            (or (env-cell use-env key)
+                (error "identifier property key must be a bound identifier" key)))
+           (value (eval (cadr (cddr expr)) use-env))
+           (this-id-props
+            (hash-table-intern! *identifier-properties*
+                                (cons use-env id-cell)
+                                (lambda ()
+                                  (make-hash-table id-comparator)))))
+      (hash-table-set! this-id-props
+                       (cons use-env key-cell)
+                       value)
+      (if #f #f))))
+
+(define (capture-lookup proc)
+  (lambda (use-env)
+    (proc
+     (letrec
+         ((f
+           (case-lambda
+             ((id key)
+              (f use-env id use-env key))
+             ((id-env id key-env key)
+              (let ((cell (env-cell id-env id)))
+                (cond ((not cell)
+                       (error "identifier not bound in environment"
+                              id id-env))
+                      ((hash-table-ref/default
+                        *identifier-properties*
+                        (cons id-env cell)
+                        #f)
+                       =>
+                       (lambda (this-id-props)
+                         (hash-table-ref/default
+                          this-id-props
+                          (cons key-env (env-cell key-env key))
+                          #f)))
+                      (else #f)))))))
+       f))))
--- a/lib/srfi/213.sld
+++ b/lib/srfi/213.sld
@@ -1,0 +1,9 @@
+(define-library (srfi 213)
+  (import (chibi)
+          (chibi ast)
+          (scheme case-lambda)
+          (scheme comparator)
+          (scheme hash-table))
+  (export define-property
+          capture-lookup)
+  (include "213.scm"))


mnieper:
>     * The implementation of `capture-lookup` is arguably not faithful to the requirement that ‘`capture-lookup` must be substitutable by the identity (lambda (proc) proc) when used in macro transformers, but I’m not really sure what that’s meant to mean. (Why have `capture-lookup` at all if it’s practically an identity function?)

This is in anticipation of future extensions to the macro system where transformers may request other expand-time values besides the `lookup` procedure, e.g. a hypothetical `frobnicate` procedure. Then we would have `capture-frobnicate` and `capture-lookup` is there for eventual symmetry. Moreover, future additions to identifier properties can add optional arguments to  `capture-lookup` that wouldn't be available when just a procedure taking one argument would be returned.

Usually, one would probably code without an explicit `capture-lookup` as it removes one of the already many indentation levels. An implementation that won't work without an explicit `capture-lookup` is probably more trouble than worth. On the other hand, adapting macros to it is mechanically simple as just a few `capture-lookup`s have to be dropped here and their in the code.

(`capture-lookup` wasn't in the original draft.)

> ‘Properties are attached to identifiers and not to bindings’ skipped because I don’t understand it (the spec document says they’re attached to bindings?)

In the sense of the spec, the `binding` is to what a particular identifier is bound (e.g. a variable, keyword, etc.). So the quoted sentence means that if, say, `+` is imported from `(scheme base)` both under the name `+` and `plus` (e.g. using the `rename` facility of the module system), using `define-property` on the imported `+` will make it visible on the imported `plus` as well.

> But given how compromised this whole thing is, especially considering this is another change to the implementations of the core macro transformer constructors (to support `capture-lookup` as the spec intends), I’m not sure if merging would actually be a good idea.

And given that the `syntax-case` implementation is also compromised, it probably makes more sense in the long run to have a version of Chibi that understands `syntax-case` and its syntactic extensions (like identifier macros) natively instead of spending too much developer's time on adding suboptimal solutions.

dpk:
> > ‘Properties are attached to identifiers and not to bindings’ skipped because I don’t understand it (the spec document says they’re attached to bindings?)
> 
> In the sense of the spec, the `binding` is to what a particular identifier is bound (e.g. a variable, keyword, etc.). So the quoted sentence means that if, say, `+` is imported from `(scheme base)` both under the name `+` and `plus` (e.g. using the `rename` facility of the module system), using `define-property` on the imported `+` will make it visible on the imported `plus` as well.

The test cases seem to be expecting the opposite, though? Unless I misunderstood what `alias` is meant to do …

dpk:
Identifier properties in SRFI 213 are attached to bindings, not to symbols. See the implementation.

SRFI 213 changes the definition of macro transformer procedures so that, if returning the result of `capture-lookup` instead of an expression, the procedure passed to `capture-lookup` is re-entered and its result is the result of macro expansion rather than the procedure itself. See the SRFI 213 spec.

mnieper:
> I know identifier properties have a long Lisp tradition but I'm not a fan of them, and don't understand why this requires changes to init-7.scm.

It's a very far way from classical Lisp symbol properties to identifier properties in the sense of SRFI 213.

mnieper:
> > > ‘Properties are attached to identifiers and not to bindings’ skipped because I don’t understand it (the spec document says they’re attached to bindings?)
> > 
> > 
> > In the sense of the spec, the `binding` is to what a particular identifier is bound (e.g. a variable, keyword, etc.). So the quoted sentence means that if, say, `+` is imported from `(scheme base)` both under the name `+` and `plus` (e.g. using the `rename` facility of the module system), using `define-property` on the imported `+` will make it visible on the imported `plus` as well.
> 
> The test cases seem to be expecting the opposite, though? Unless I misunderstood what `alias` is meant to do …
> 
> EDIT: Okay, now I’m really confused. Unsyntax:
> 
> ```scheme
> #,> (import (rename (only (scheme base) +) (+ plus)))
> #,> (define foo "useless value")   
> #,> (define-property + foo "useful value")
> #,> (identifier-property + foo)
> "useful value"
> #,> (identifier-property plus foo)
> #f
> ```
> 
> (Docker image `schemers/unsyntax:latest`/Unsyntax 0.0.3.25-1447-dirty)
> 
> Is this a bug in Unsyntax that wound its way into the test suite as specified behaviour? Or am I seriously misunderstanding?
> 
> EDIT 2: Okay, Chez does the same. So I’m seriously misunderstanding. What makes the bindings of `+` and `plus` ‘distinct’ in this example even though they’re `free-identifier=?`? Do identifiers have to be both `bound-identifier=?`/`symbolic-identifier=?` and `free-identifier=?` to have the same identifier properties? (No, wait, that wouldn’t make sense either …)

Can you test the above as part of a program and not at the REPL? 

Can you also cut and paste the test you think clashes with the definition of `alias`?

dpk:
<details>
<summary>Chez Scheme version</summary>

```scheme
(import (rnrs (6))
        (rename (only (rnrs (6)) +) (+ plus))
        (only (chezscheme) define-property))


(define-syntax identifier-property
  (lambda (stx)
    (lambda (lookup)
      (syntax-case stx ()
        ((_ identifier key)
         (with-syntax ((value (lookup #'identifier #'key)))
           #'(quote value)))))))

(define foo "useless value")
(define-property + foo "useful value")

(write (identifier-property + foo))
(newline)
(write (identifier-property plus foo))
(newline)
```

</details>

<details>
<summary>Unsyntax version</summary>

```scheme
(import (scheme base)
        (rename (only (scheme base) +) (+ plus))
        (scheme write)
        (srfi 211 syntax-case)
        (srfi 213))


(define-syntax identifier-property
  (lambda (stx)
    (capture-lookup
     (lambda (lookup)
       (syntax-case stx ()
         ((_ identifier key)
          (with-syntax ((value (lookup #'identifier #'key)))
            #'(quote value))))))))

(define foo "useless value")
(define-property + foo "useful value")

(write (identifier-property + foo))
(newline)
(write (identifier-property plus foo))
(newline)
```

</details>

Output in both Racket-Chez 9.5.5.6 (`scheme --program file.scm`) and Unsyntax (version from Docker as above, `unsyntax-compile file.scm`, `./file`):

```
"useful value"
#f
```

I meant [the last two tests in the file](https://github.com/scheme-requests-for-implementation/srfi-213/blob/master/tests/srfi-213-tests.sch#L169) but their behaviour in Unsyntax is seemingly consistent with that of renaming imports. So it’s not `alias` in particular but aliasing/renaming in general that’s confusing me.

dpk:
Okay, I think I understand what’s going on. I think that in both Chez and Unsyntax, when a binding is imported or aliased, a copy of its identifier property store is made and this new copy is associated with the name, instead of a single identifier property store per actual binding. That’s consistent at least with the behaviour shown by Chez here:

```scheme
(library (test-library)
  (export test-identifier test-key)
  (import (chezscheme))

  (define test-identifier 1)
  (define test-key 2)
  
  (define-property test-identifier test-key 3))
```

```scheme
> (import (test-library)
          (rename (only (test-library) test-identifier)
                  (test-identifier another-test-identifier)))
> (define-syntax identifier-property
    (lambda (stx)
      (lambda (lookup)
        (syntax-case stx ()
          ((_ identifier key)
           (with-syntax ((value (lookup #'identifier #'key)))
             #'(quote value)))))))
> (identifier-property test-identifier test-key)
3
> (identifier-property another-test-identifier test-key)
3
> (define-property another-test-identifier test-key 4)
> (identifier-property another-test-identifier test-key)
4
> (identifier-property test-identifier test-key)
3
```

This doesn’t seem to be explicitly mentioned in SRFI 213 to me, but perhaps it’s implied by something I’m not quite reading correctly. It also seems, conceptually, like the Wrong Thing, though I may be biased by just having implemented a version of SRFI 213 that does this differently.

mnieper:
> In the sense of the spec, the `binding` is to what a particular identifier is bound (e.g. a variable, keyword, etc.). So the quoted sentence means that if, say, `+` is imported from `(scheme base)` both under the name `+` and `plus` (e.g. using the `rename` facility of the module system), using `define-property` on the imported `+` will make it visible on the imported `plus` as well.

My mistake. I shouldn't answer to Git* issues when I am doing something entirely else at the same time.

You correctly wrote above that "properties are attached to identifiers and not to bindings" according to the spec, but when interpreting that sentence this morning I somehow managed to turn it around in my head.

So the spec is correct as is Chez's and Unsyntax's behavior with respect to the spec. Renaming an imported identifier or aliasing one does not introduce a new binding but that's irrelevant to identifier properties because they are attached to the name, not its meaning.

Thus, the relevant equivalence predicate is `bound-identifier=?`, not `free-identifier=?`.

Sorry again for the confusion.

mnieper:
> In that case, I find the wording used in SRFI 213 extremely misleading. 

Agreed. I will choose clearer wording in the Foundations document.

> I implemented it here assuming ‘binding’ meant the same as it does in the definition of `free-identifier=?`. I see nowhere in SRFI 213 that suggests that a different sense of the word ‘binding’ is being used.

The problem of the word sense becomes only apparent in conjunction with renames/aliases: From one point of view, `define-property` modifies an existing binding and `rename`/`alias` guard against it by copying the binding. It is always the same problem: What is meant by "the same" when mutation is present. Anyway, as said above, the wording will be improved when both identifier properties and aliases appear in the same document.

In any case, it cannot be `free-identifier=?` because this would break the lexical scoping of `define-property`.

> This is not my only complaint with SRFI 213: there’s no way to tell when a property is undefined vs when it has been defined to `#f`, for example.

Acknowledged. The Foundation version of the lookup procedure will allow a third, optional argument taking a default value.

(It's not a serious disadvantage in practice, but being able to choose an arbitrary default value makes it more complete.)

PR #821, opened 2022-04-09
(chibi parse): Ensure reason is always a string

nmeum:
While testing a `(chibi parser)`-based parser I noticed that
`parse-string` is the only provided parser combinator which uses a list,
instead of a string, as a failure reason. It is not explicitly
documented but since all other standard parser combinator use a string
for the error reason I assume this to be a bug and have adjusted the
`parse-string` combinator accordingly in this commit.


ashinn:
Can you instead change parse-with-failure-reason to lazily check if the argument is a list and string-concatenate it before use?  We don't want to pay for this if not needed.

ashinn:
Ah, nm, that would push the construction from construction time to parse time.

PR #830, opened 2022-05-11
allow user-specified CHIBI chibi-scheme path

stevengj:
I'm currently attempting to cross-compile chibi-scheme (https://github.com/JuliaPackaging/Yggdrasil/pull/4894), but in this situation you cannot use the built `chibi-scheme` executable (which is for the target architecture) to build image files, since the image-file creation runs on the host architecture.

To work around this, this PR allows you to `cmake -DCHIBI=...` to specify a path to `chibi-scheme` for bootstrapping (e.g. by first building on the host architecture as usual, and then starting another build to cross-compile).

PS. Any other advice about cross-compiling chibi-scheme would be welcome.  For example, are the image files portable between architectures or is there some special consideration when cross-compiling them?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,7 +207,7 @@ else()
     set(bootstrap chibi-scheme-bootstrap)
 endif()
 
-# when cross-compiling, we may nee
+# when cross-compiling, we need to use a separately built chibi-scheme executable:
 set(BOOTSTRAP ${bootstrap} CACHE FILEPATH "chibi-scheme path for bootstrapping")
 
 function(add_stubs_library stub)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,7 +208,7 @@ else()
 endif()
 
 # when cross-compiling, we need to use a separately built chibi-scheme executable:
-set(BOOTSTRAP ${bootstrap} CACHE FILEPATH "chibi-scheme path for bootstrapping")
+set(CHIBI ${bootstrap} CACHE FILEPATH "chibi-scheme path for bootstrapping")
 
 function(add_stubs_library stub)
     set(link-libraries LINK_LIBRARIES)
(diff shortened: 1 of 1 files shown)


ashinn:
Note the Makefile already supports this with the CHIBI variable.  The CMake build is only unofficially supported.

Images won't work if the target platform has different configurations.  They might work otherwise but I wouldn't guarantee it.

stevengj:
Thanks, I can try that, but my concern was that the `Makefile` might not be set up for cross compilation?   But hopefully if I set up `CC`, `LD`, etcetera it should work.

The host (build system) and target system can use `chibi-scheme` builds with identical configurations (though possibly completely different CPU architectures etc.)

stevengj:
I've renamed `BOOTSTRAP` to `CHIBI` to match the `Makefile`

stevengj:
Closing since you prefer the Makefile; I'll work on cross-compiling with that.

ashinn:
Improvements to the CMake build are welcome, I just don't maintain it myself.

ashinn:
Note the emscripten build is cross-platform.  In general cross-platform builds should just work, however the image handling is tricky - this is technically part of the installation process, not the build, and has to be done on the final target.  On the other hand images are optional to begin with.

stevengj:
(Thanks for the note about the emscripten build.  I'd like a native cross-platform build since I'd like to link to `libchibi-scheme` from Julia using its FFI.)

Cross-compiling for Windows is a bit tricky because `chibi-ffi lib/chibi/filesystem.stub` is choosing the wrong in `(cond-expand (windows ...))`.   It seems like judicious use of `-Dwindows`, or more generally `(set-cdr! (features) '(...))` might fix this, and perhaps fix image file generation (though it is nice to know that I can omit the images if needed).

ashinn:
Good point about the FFI - let me add an option to override the platform in chibi-ffi.

stevengj:
> Good point about the FFI - let me add an option to override the platform in chibi-ffi.

Isn't the `--features` flag supposed to do this already?   But I tried
```sh
chibi-ffi --features chibi,r7rs,ratios,complex,uvector,threads,full-unicode,modules,dynamic-loading,windows lib/chibi/filesystem.stub
```
and it doesn't seem to generate the Windows version of `filesystem.c`.  What is missing?

PR #832, opened 2022-05-14
SRFI-144: accept zero arguments for flmax/flmin

jpellegrini:
SRFI-144 requires that `(flmin)` returns +inf.0 and that `(flmax)` returns -inf.0 when infinities are supported, so these procedures can't really be aliases to the Chibi implementation of R7RS max and min.


ashinn:
Could you add a unit test while you're at it?  Thanks!

jpellegrini:
> Could you add a unit test while you're at it? Thanks!

Sure, just added some.

PR #843, opened 2022-07-29
Add (chibi shell) to the documentation.

lockywolf:
1. Add "shell" to the list of html_docs
2. Add a reference to (chibi shell) to chibi.scribl

I would like to also ask to regenerate the docs at http://synthcode.com/scheme/chibi/
(Perhaps it would not be too much of a burden to move the manual to Github Pages? Then the manual can be automatically regenerated on every push.)

### DIFF
--- a/doc/chibi.scrbl
+++ b/doc/chibi.scrbl
@@ -1374,6 +1374,8 @@ namespace.
 
 \item{\hyperlink["lib/chibi/repl.html"]{(chibi repl) - A full-featured Read/Eval/Print Loop}}
 
+\item{\hyperlink["lib/chibi/shell.html"]{(chibi shell) - Binding to the most necessary POSIX system calls.}}
+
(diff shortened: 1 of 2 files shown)


doc/chibi.scrbl
```diff
@@ -1374,6 +1374,8 @@ namespace.
 
 \item{\hyperlink["lib/chibi/repl.html"]{(chibi repl) - A full-featured Read/Eval/Print Loop}}
 
+\item{\hyperlink["lib/chibi/shell.html"]{(chibi shell) - Binding to the most necessary POSIX system calls.}}
```
ashinn, line 1377: Sorry, I merged before noticing, but this actually doesn't export any system calls or have anything to do with POSIX.

I pushed a new description.

lockywolf: My bad, that description wasn't too accurate.

But would you consider hosting documentation on "Github Pages" re-generating on push?
I suspect it would amount to less work than publishing it manually.

PR #847, opened 2022-08-02
Install manpages in man/man1, not man

lockywolf:
Shouldn't man pages be installed in man/man1?


lockywolf:
The appveyor failure is unrelated to the man pages installation.

PR #858, opened 2022-09-11
cmake: Exclude `(chibi shell)` test on Win32

okuoku:
Exclude `(chibi shell)` test on Win32 since it's not compatible.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -410,6 +410,7 @@ set(win32testexcludes
     chibi/tar-test # Depends (chibi system)
     chibi/process-test # Not applicable
     chibi/pty-test # Depends (chibi pty)
+    chibi/shell-test # Depends Linux procfs
     )
 
 foreach(e ${srfi_tests} ${chibi_scheme_tests})


okuoku:
Re: `(chibi shell)`:

- `/proc/self/fd` is Linux specific. 
- Perhaps you meant `/dev/fd` here? https://github.com/ashinn/chibi-scheme/blob/0a50b305bc508003263e764949c25b9a21cccf91/lib/chibi/shell.scm#L199

ashinn:
> Re: `(chibi shell)`:
> 
>     * `/proc/self/fd` is Linux specific.  (EDIT: Some OSes provides compatibility layer: Cygwin provide it by-default, FreeBSD have linuxprocfs, ...)
> 
>     * Perhaps you meant `/dev/fd` here? https://github.com/ashinn/chibi-scheme/blob/0a50b305bc508003263e764949c25b9a21cccf91/lib/chibi/shell.scm#L199

They both work.

ashinn:
Thanks!

okuoku:
> Perhaps you meant `/dev/fd` here? 

(Solved in #861 )

ashinn:
Sorry, the way it was written I thought you meant `/dev/fd` instead of `/proc/self/fd`.

PR #867, opened 2022-10-10
Add support for SXML indentation on output.

arthurgleckler:
This adds a second optional argument to `sxml-display-as-html`.  If true, it adds newlines and indentation where appropriate.  This makes reading the HTML easier.


lassik:
We should probably use keyword arguments or a settings object of some kind. It's not easy to predict what options will be added in the future.

lib/chibi/sxml.scm
```diff
@@ -80,13 +80,24 @@
...
 ;;> \var{@raw} tag is considered safe text and not processed or escaped.
 (define (sxml-display-as-html sxml . o)
-  (let ((out (if (pair? o) (car o) (current-output-port))))
+  (let-optionals o ((out (current-output-port))
+		    (indent? #false))
     (let lp ((sxml (if (and (pair? sxml) (eq? '*TOP* (car sxml)))
                        (cdr sxml)
-                       sxml)))
+                       sxml))
+	     (i 0))
```
ashinn, line 100: nit: i sounds like a normal counter, maybe better to name this depth or indent?

ashinn (approved):
Two optional arguments is fine, no need to use keywords here.

lib/chibi/sxml.scm
arthurgleckler: Good idea.  I renamed it `depth`.

PR #868, opened 2022-10-12
Fix bug: attributes without values didn't work.

arthurgleckler:
Before, it was necessary to do something like this:

  (option (@ (selected . #false) (value "any")) "any")

instead of:

  (option (@ (selected) (value "any")) "any")

Only the former is valid SXML, as far as I can tell from the SXML specification:

  <https://dl.acm.org/doi/pdf/10.1145/571727.571736>

### DIFF
--- a/lib/chibi/sxml.scm
+++ b/lib/chibi/sxml.scm
@@ -40,11 +40,11 @@
     (lambda (out) (html-display-escaped-attr (display-to-string str) out))))
 
 (define (html-attr->string attr)
-  (if (cdr attr)
+  (if (null? (cdr attr))
+      (symbol->string (car attr))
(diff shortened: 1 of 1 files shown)


ashinn:
I'm not sure if I deliberately changed this, or if I searched, found the link you shared (which I can't access), and gave up (the free spec is at https://okmij.org/ftp/Scheme/SXML.html).

Either way it was a mistake :(

We'll have to be permissive in what we accept and try to fix the places we generate SXML later.

ashinn:
Disregard my comment, I thought from your description that we were using dotted lists for attrs, but this code already supported both (just not an empty value), and libraries that generate SXML like `(chibi html-parser)` are correctly using proper lists.

PR #870, opened 2022-10-26
Make macro-aux safe for other things together with syntax-case

dpk:
If you set the macro-aux of a macro outside of (chibi syntax-case), it would previously case `syntax` to think that it was a pattern variable and try to substitute it, even if the macro-aux was being used for something else.

This patch fixes that by wrapping pattern variable values in an extra typed box and checking that it has the right type before deciding that it’s actually a pattern variable.

### DIFF
--- a/lib/chibi/syntax-case.scm
+++ b/lib/chibi/syntax-case.scm
@@ -61,6 +61,10 @@
     ((letrec-syntax ((keyword transformer) ...) . body)
      (%letrec-syntax ((keyword (make-transformer transformer)) ...) . body))))
 
+(define-record-type Pattern-Cell
+  (make-pattern-cell val) pattern-cell?
+  (val pattern-cell-value))
+
 (define-syntax define-pattern-variable
   (er-macro-transformer
    (lambda (expr rename compare)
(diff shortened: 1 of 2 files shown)


mnieper:
> How can the aux field be used to implement SRFI 213?

I thought I would just drop this random idea and then someone else would figure out whether it would work. :smiley: 

I had in mind that when `define-property` is used to add a property to an identifier, the identifier is rebound to a macro that is "identifier syntax".  Then, the aux field of the macro is used for meta-information.  This would need some amendment to `free-identifier?` because the meaning of the rebound identifier as tested by this relation shouldn't change.

But I haven't thoroughly thought about it so expect the devil in the details

mnieper:
> Yes, my initial thought was you must mean something like that, but it would have the same problem as my approach in #820, no? (I assume this is what you mean with the problem with `free-identifier=?`)

I didn't mean that problem (but that is probably still a problem). I just mean that redefining the binding of an identifier (from a variable to a keyword) would break `free-identifier=?` if this rebinding by `define-property` is not understood and ignored by `free-identifier=?`.

So it may be better to attach an aux field to every kind of bound object, not only macros.

PR #871, opened 2022-10-28
Fix syntax-tests to use mutable-environment

dpk:
The AppVeyor cmake build was failing because the identifier syntax tests used `environment` assuming its result was mutable; this broke as of 0eeeac765093a55234cbd1fdd2e04d6f20ff1027.

This patch also adds the syntax-tests to the `test-all` suite, so similar regressions in future will be caught by all the CI builds failing, not just the AppVeyor one.


ashinn:
Thanks!

Issue #885, opened 2023-01-30
Implement delimited continuations


mnieper:
Since yesterday, I have been working on a SRFI 226-compatible implementation of delimited continuations for Chibi Scheme (see here for the current development: https://github.com/mnieper/chibi-scheme/tree/srfi-226).

It basically replaces the call/cc & dynamic-wind implementation in init-7.scm so that delimited continuations can be supported.

After I have added thread support (similar to how the old implementation used the `%dk` parameter), the implementation should be ready to merge.

For efficiency reasons, it would be good if a continuation could be captured early when (a) a program is run or (b) an expression in the REPL is evaluated.  Calling this continuation with a thunk should call this thunk and deliver its values to the program runner or expression evaluator. @ashinn, can you point out to me where it is best to capture such a continuation?

Thanks,

Marc

ashinn:
Thanks for working on this!

The start for running programs is in main.c (where it says "load the module or script").

In the same file is the `repl` function, but perhaps more important is `(chibi repl)`.
The latter runs each expression in a separate thread so it can be terminated.

mnieper:
Thank you for the immediate response. After having thought about it, I have implemented a new opcode, `SEXP_OP_ABORT` that clears the stack and then executes a thunk in this empty stack.  This may also be useful in other cases to save stack when stack frames become essentially dead.

mnieper:
The implemented API is in `(srfi 226 prompt)`, `(srfi 226 continuation)` and in `(srfi 226 shift-reset)`.

Risto-Stevcev:
In case anyone reading wants to use delimited continuations right away, I have this snippet I made based on some scheme implementation I found online somewhere, it seems to work:

```
(import (scheme base)
        (scheme write)
        (scheme eval)
        (chibi show)
        (chibi match)
        (chibi test))

(define *meta-continuation*
        (lambda (value)
          (error "No top-level RESET" value)))

(define-syntax reset
  (syntax-rules ()
    ((reset body)
     (let ((mc *meta-continuation*))
       (call-with-current-continuation
         (lambda (k)
           (set! *meta-continuation*
                 (lambda (value)
                   (set! *meta-continuation* mc)
                   (k value)))
           (let ((result body))
             ;** do not beta-substitute!!
             (*meta-continuation* result))))))))

(define-syntax shift
  (syntax-rules ()
    ((shift var body)
     (call-with-current-continuation
       (lambda (k)
[...]

PR #886, opened 2023-01-30
Provide delimited continuations

mnieper:
Implement delimited continuations on top of the existing primitive %call/cc. Space leaks are prevented through a new primitive %abort.

See issue #885.


ashinn:
Thanks for working on this, this is awesome!

I guess a more optimal implementation would use continuation marks (in Chibi basically a pointer to a stack object and an offset) and slice only up until that mark, but this is OK for now.

My primary concern is the huge amount of code added to init-7.scm, which is supposed to be tiny.  How much of this can be moved into SRFI 226?

mnieper:
> Thanks for working on this, this is awesome!

Thank you! Chibi may not be the fastest implementation on the planet, but thanks to the clear code, hacking and improving it is fun.

> I guess a more optimal implementation would use continuation marks (in Chibi basically a pointer to a stack object and an offset) and slice only up until that mark, but this is OK for now.

As Chibi's `call/cc` primitive copies the whole stack, the code for the `dynamic-wind`-aware `call-with-current-continuation` in my patch may actually be an efficiency improvement because the call stack is effectively cut in slices.

What would be an improvement if Chibi's VM offered a primitive to call a thunk *in* a given continuation. Once this is in place, I could eliminate the `%call-with-current-continuation` hack.

> My primary concern is the huge amount of code added to init-7.scm, which is supposed to be tiny. How much of this can be moved into SRFI 226?

The previous `init-7.scm` exported `dynamic-wind` (and a `dynamic-wind`-aware `call/cc`). As I didn't want to remove this, I had to put substantial parts into `init-7.scm`.  If you really want to keep it small, removing `dynamic-wind` and implementing `call-with-current-continuation` in `(srfi 226)` seems the way to go. (The library `(scheme base)` can then import `call-with-current-continuation` from `(srfi  226)` and not from `(chibi)`.)

ashinn:
Hmm... the ctak benchmark is almost 4x slower with this.  Currently we copy the whole stack, but in practice the stack is rarely large, and the copy is just an optimized C loop (actually we could tune it more).

mnieper:
> Hmm... the ctak benchmark is almost 4x slower with this. Currently we copy the whole stack, but in practice the stack is rarely large, and the copy is just an optimized C loop (actually we could tune it more).

Some of the slow-down is because of the `%call-with-current-continuation` hack that is currently necessary because there is no primitive (VM instruction) calling a thunk in a continuation (i.e. throwing away the current continuation, replacing it by the given one and then calling the thunk given this new call stack. Adding such a VM instruction looks simple.

The new code in `init-7.scm` was written for clarity and correctness (and to mimic the code in the sample implementation of SRFI 226).  One can probably fine-tune it and expand procedures/records by hand so that Chibi is able to run it faster.

ashinn:
But this exists - it's `SEXP_OP_RESUMECC`.

mnieper:
> But this exists - it's `SEXP_OP_RESUMECC`.

Good. What are the parameters that this bytecode expects? And can it be called directly from the user code?

PR #887, opened 2023-02-01
Provide identifier-syntax and make-variable-transformer through standardized SRFI libraries.

mnieper:
To make writing portable code easier, SRFI 211 defines the libraries `(scheme 211 identifier-syntax)` and `(scheme 211 variable-transformer)` that export `identifier-syntax` and `make-variable-transformer`, respectively.

This change defines these two libraries in Chibi.


ashinn:
Thanks!

Issue #882, opened 2023-01-13
sexp_destroy_context never returns during shutdown


mars2klb:
When exiting, it seems that `sexp_destroy_context` never returns.

On linux, this _appears_ to be ok, because the exit code is 0. But on plan9 this causes failures because `exits` is not called correctly, so the exit is considered a failure.

For testing, I just put some print statements in `main.c` (around the function calls near line 658)
Tested on 9front and Void Linux.

All I do is go into the repl, then exit. Messages are emitted up to `main.c:658`, but the process exits before the rest is executed.

If I enable `SEXP_USE_GLOBAL_HEAP`, messages _after_ are emitted as expected.

ashinn:
On Linux, OSX and other regularly used platforms, of course `sexp_destroy_context` returns, otherwise we're either crashing or looping.  So if it doesn't return, this would likely be from some memory corruption or misalignment in Plan9 causing the gc to loop.

There is no code path calling `exit` from `sexp_destroy_context`.

ashinn:
Ah, you're probably being confused trying to use printf debugging here instead of an external debugger.

`sexp_destroy_context` finalizes everything, including all open ports, so after it returns stdout and stderr are closed.

mars2klb:
You are 100% correct! Sorry about the noise.

PR #888, opened 2023-02-04
Fix exit call on plan9/9front

smazga:
* make sure exits is passed a nil on success (a zero isn't good enough)
* make sure main exits with a proper exits call

This addresses the issue that lead to the erroneous issue #882

### DIFF
--- a/main.c
+++ b/main.c
@@ -684,11 +684,15 @@ int main (int argc, char **argv) {
   res = run_main(argc, argv);
   if (sexp_fixnump(res)) {
     int code = sexp_unbox_fixnum(res);
+#ifdef PLAN9
     if (code == 0) {
       exit_success();
     } else {
       exit_failure();
     }
+#else
+    return code;
+#endif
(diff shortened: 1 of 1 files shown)


main.c
```diff
@@ -682,7 +683,12 @@ int main (int argc, char **argv) {
   sexp_scheme_init();
   res = run_main(argc, argv);
   if (sexp_fixnump(res)) {
-    return sexp_unbox_fixnum(res);
+    int code = sexp_unbox_fixnum(res);
```
ashinn, line 686: Can you ifdef this for Plan9, so users on other OSes can still specify the exit code?

ashinn (changes requested):

main.c
smazga: How does this option work?

That logic could probably also be abstracted away so it doesn't clutter up `main()`

ashinn:
Thanks!

PR #894, opened 2023-02-15
Use the Scheme value of void for json null instead of the Scheme null value

### LINKED ISSUE #893: json-write cannot write null
```scheme
> (eq? (if #f #f) (string->json "null"))
#t
> (json->string (string->json "null"))
ERROR in "json-write": unable to encode element: #<undef>
```

`json->string` (and the underlying `json-write`) should write the Scheme representation of the JSON null as `null`.

Zambito1:
Closes #893.


Zambito1:
Fixed in https://github.com/ashinn/chibi-scheme/commit/c026c0884ddc1512d09aa5cde3fd284f9b98f614

PR #903, opened 2023-03-15
add simple http client and server examples

adamfeuer:
Adds simple http client and server examples:

- http client can get the contents of a URL specified on the command line; or print usage if no URL is givien.
- http server serves a very simple HTML page with a single number that is incremented on every request. (Adapted from the source code documentation.)

### DIFF
--- a/examples/simple-http-client.scm
+++ b/examples/simple-http-client.scm
@@ -1,0 +1,36 @@
+#! /usr/bin/env chibi-scheme
+
+; Simple HTTP client 
+; Retrieves the contents of the URL argument:
+
+; Usage:
+; simple-http-client.scm [URL]
+;
+; Example:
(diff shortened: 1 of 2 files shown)


examples/simple-http-client.scm
```diff
@@ -0,0 +1,36 @@
...
+; Example:
+; simple-http-client.scm http://localhost:8000
+
+(import (chibi) (chibi net) (chibi net http) (chibi io))
+
+(if (> (length (command-line)) 1) 
+  (let ((url (car (cdr (command-line)))))
+        (if (> (string-length url) 0) 
+          (begin
+            (display (read-string 65536 (http-get url)))
```
ashinn, line 18: Note you can use port->string here.

ashinn (approved):
Thanks!

PR #907, opened 2023-03-31
add support for Guile [snow-chibi]

rgherdt:
Two recent changes to Guile improved its R7RS support:

- 3.0.7 fixed a bug regarding cond-expand in a define-library form.
- 3.0.8 added support for R7RS' srfi library names, e.g. (srfi 69) instead of (srfi srfi-69) used by Guile.

These changes open the possibility for using snow libraries in R7RS Guile programs. This commit adds support for installing/removing snow libraries for Guile.


ashinn:
Thanks!

PR #913, opened 2023-04-21
Drop (chibi sxml) dependency on let-optionals

lassik:
let-optionals can be provided by the (chibi optional) library or by SRFI 227. Either dependency is non-trivial and makes it tricky to incorporate (chibi sxml) into outside projects. Since (chibi sxml) only makes trivial use of let-optionals, expand the macro by hand.

### DIFF
--- a/lib/chibi/sxml.scm
+++ b/lib/chibi/sxml.scm
@@ -91,9 +91,12 @@
 
 ;;> Render (valid, expanded) \var{sxml} as html.
 ;;> \var{@raw} tag is considered safe text and not processed or escaped.
-(define (sxml-display-as-html sxml . o)
-  (let-optionals o ((out (current-output-port))
-		    (indent? #false))
+(define (sxml-display-as-html sxml . args)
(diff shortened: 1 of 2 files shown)


lib/chibi/sxml.scm
```diff
@@ -91,9 +91,15 @@
 
 ;;> Render (valid, expanded) \var{sxml} as html.
 ;;> \var{@raw} tag is considered safe text and not processed or escaped.
-(define (sxml-display-as-html sxml . o)
-  (let-optionals o ((out (current-output-port))
-		    (indent? #false))
+(define (sxml-display-as-html . args)
+  (when (null? args) (error "not enough args"))
+  (let* ((sxml (car args))
```
ashinn, line 96: You can simplify this as:

```
(define (sxml-display-as-html sxml . args)
  ...)
```

ashinn (changes requested):
Thanks!

lib/chibi/sxml.scm
lassik: lol, true

PR #915, opened 2023-05-04
Fix broken import in (srfi 193)

### LINKED ISSUE #914: Can't import srfi 193
Using a freshly cloned repository:

```
> (import (srfi 193))
ERROR: importing unknown binding
    command-line
    (*this-module* *this-path* make-module %module-exports module-env module-env-set! module-meta-data module-meta-data-set! module-exports module-name->strings module-name->file module-name-prefix load-module-definition find-module *modules* add-module! delete-module! symbol-append symbol-drop warn to-id from-id id-filter %resolve-import auto-generate-bindings *chibi-env* *auto-env* resolve-import resolve-module-imports load-module eval-module mutable-environment environment meta-begin meta-define define-library-transformer define-library module pop-this-path include-library-declarations define-meta-primitive import import-immutable export export-all include include-ci include-shared include-shared-optionally body begin repl-import raw-script-file)
```

Works with the 0.10.0 release.

lassik:
Fixes #914

### DIFF
--- a/lib/srfi/193.sld
+++ b/lib/srfi/193.sld
@@ -2,7 +2,8 @@
 (define-library (srfi 193)
   (export command-line command-name command-args script-file script-directory)
   (import (scheme base) (chibi filesystem) (chibi pathname)
-          (only (meta) command-line raw-script-file))
+          (only (chibi) command-line)
+          (only (meta) raw-script-file))
   (begin
 
     (define (command-name)


ashinn:
Thanks!

shawnw:
Doesn't fix the entire problem; see https://github.com/ashinn/chibi-scheme/issues/914#issuecomment-1535721048

PR #918, opened 2023-05-23
Add index-swap, fix assert in array-append

gambiteer:
These changes are untested.


ashinn (approved):
Thanks!

gambiteer:
I noticed that some arguments were protected by asserts, and others were not.
Do you want argument checks for all arguments?

ashinn:
Yes, in general there should be more checks.  This is still a WIP but I hope to wrap up soon.  Notably `array-block` is not implemented, and the signature of make-specialized-array is incorrect (and probably a few other details).

PR #937, opened 2023-08-28
SRFI 231: Declare char-storage-class, fix interval-projections

gambiteer:


gambiteer:
If you have time for pull requests, I may work on some more tests and small fixes.

ashinn (approved):
Thanks!

gambiteer:
I don't think I can commit to the chibi repository.

PR #939, opened 2023-09-27
Handle basic special characters in `write`

raviqqe:
This PR fixes the implementation of `write` in SRFI 38 with special characters (e.g. newline and tab) as described in the section "6.13.3. Output" in the [R7RS](https://small.r7rs.org/attachment/r7rs.pdf).


raviqqe:
Ah, it looks like it's not necessary to escape them with their names but we can put any character after `#\`. I'm gonna just close this.

ashinn:
We should  write the descriptive names.  We can either use this list or punt to the underlying write like we do for numbers.

ashinn (approved):

PR #940, opened 2023-10-13
Unwind on exit

raviqqe:
# Description

This PR implements the unwinding of dynamic extents on a call to the `exit` procedure as described in "6.14. System interface" in [the R7RS specification](https://small.r7rs.org/attachment/r7rs.pdf) using `call/cc`.

# Questions

- [ ] Where can I write tests for the new behaviour with side effects of `exit`?
- [x] When I build the library, it emits those warnings. But I don't know why it happens while I'm still using the `only` predicate in imports.
 
```sh
> make install
...
if type ldconfig >/dev/null 2>/dev/null; then ldconfig; fi
echo "Generating images"
Generating images
[ -z "" ] && LD_LIBRARY_PATH="local/lib:" DYLD_LIBRARY_PATH="local/lib:" CHIBI_MODULE_PATH="local/share/chibi:local/lib/chibi" local/bin/chibi-scheme -mchibi.repl -d local/share/chibi/chibi.img
[ -z "" ] && LD_LIBRARY_PATH="local/lib:" DYLD_LIBRARY_PATH="local/lib:" CHIBI_MODULE_PATH="local/share/chibi:local/lib/chibi" local/bin/chibi-scheme -xscheme.red -mchibi.repl -d local/share/chibi/red.img
[ -z "" ] && LD_LIBRARY_PATH="local/lib:" DYLD_LIBRARY_PATH="local/lib:" CHIBI_MODULE_PATH="local/share/chibi:local/lib/chibi" local/bin/chibi-scheme -mchibi.snow.commands -mchibi.snow.interface -mchibi.snow.package -mchibi.snow.utils -d local/share/chibi/snow.img
WARNING: importing already defined binding: exit
WARNING: importing already defined binding: exit
```

Thanks ahead!

### DIFF
--- a/lib/chibi/process.scm
+++ b/lib/chibi/process.scm
@@ -1,19 +1,28 @@
+(define unwind #f)
+
+((call/cc
+    (lambda (k)
+      (set! unwind k)
+      (lambda () #f))))
 
 (cond-expand
  (plan9
-  (define (exit . o)
+  (define (emergency-exit . o)
     (%exit (if (pair? o)
                (if (string? (car o))
                    (car o)
                    (if (eq? #t (car o)) "" "chibi error"))
                ""))))
  (else
-  (define (exit . o)
+  (define (emergency-exit . o)
     (%exit (if (pair? o)
                (if (integer? (car o))
                    (inexact->exact (car o))
                    (if (eq? #t (car o)) 0 1))
                0)))))
+
+(define (exit . o)
+  (unwind (lambda () (apply emergency-exit o))))
 
 (cond-expand
  (bsd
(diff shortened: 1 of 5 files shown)


ashinn:
Sorry for the long delay.  I believe the warnings are from libraries which import both `(chibi proces)` and `(scheme process-context)`, which both export `exit`.  Before your change, this was the same binding so there was no warning.

I think the cleaner fix is to move these changes to `(chibi process)` which can export both `exit` and `emergency-exit` and just re-export these from `(scheme process-context)`.

ashinn:
Thanks!

raviqqe:
Thank you too!

PR #942, opened 2023-11-06
`(chibi test)`: Put newlines after test results etc in verbose mode

dpk:


ashinn (approved):
Thanks!  I haven't been using verbose mode but will take your word for it :)

PR #945, opened 2024-01-08
Fix #944: concatenate! work with empty lists in any position 

### LINKED ISSUE #944: SRFI-1 `append!` and `append` not equivalent
Hi,

I noticed `append!` doesn't work when the first list is empty and that is not equivalent to what `append` does, which just works.

I think there's some bug in `append!`'s implementation, that relies on the first element having a `cdr` which it may not have.

I'll try to find a fix and propose it.

Thanks!

ekaitz-zarraga:
Fixes #944


lib/srfi/1/misc.scm
```diff
@@ -15,12 +15,13 @@
 (define (concatenate! lists)
   (if (null? lists)
       '()
-      (let lp ((ls lists))
-        (cond ((not (pair? (cdr ls)))
-               (car lists))
-              (else
-               (set-cdr! (last-pair (car ls)) (cadr ls))
-               (lp (cdr ls)))))))
+      (fold (lambda (el acc)
```
ashinn, line 18: Can you add a unit test for this case?

ashinn (commented):
Thanks!

lib/srfi/1/misc.scm
ekaitz-zarraga: Added in 70989e0cef3ccb83d7391582c1d0ab0459bb1954

ashinn (approved):
Thanks!

PR #946, opened 2024-01-09
Reduce iterations in concatenate!

ekaitz-zarraga:
This commit should reduce the amount of iterations in concatenate to N where N is the sum of the lengths of the input lists.

The previous implementation iterated from the beginning in each concatenation because of `last-pair`.

This implementation is significantly faster in this extreme case:

``` scheme
(concatenate! `(,(iota 50000) ,@(map list (iota 500))))
```

Previous implementation:
real	0m0.671s
user	0m0.658s
sys	0m0.013s

This implementation:
real	0m0.175s
user	0m0.174s
sys	0m0.001s

The tests is done using `time`, which is not reliable at all, but using `(trace last-pair)` shows accurately what happens with the iterations.

### DIFF
--- a/lib/srfi/1/misc.scm
+++ b/lib/srfi/1/misc.scm
@@ -15,13 +15,16 @@
 (define (concatenate! lists)
   (if (null? lists)
       '()
-      (fold (lambda (el acc)
-                (cond
-                  ((null? acc) el)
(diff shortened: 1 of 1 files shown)


ashinn (approved):
Thanks!

PR #950, opened 2024-03-17
Avoid needless allocation in read-bytevector!

wasamasa:
This change switches the implementation strategy to basing read-bytevector on top of read-bytevector! rather than the other way around. Bonus: It's a bit less code than previously and is less surprising to the reader.


ashinn (changes requested):
Thanks!  I'm not sure why I did it this way - it should really be done in the FFI but your approach is better until I get around to that.

lib/scheme/extras.scm
```diff
@@ -136,35 +136,28 @@
...
-        (let lp ((i 0))
-          (if (>= i n)
-              res
-              (let ((x (read-u8 in)))
-                (cond ((eof-object? x)
-                       (if (zero? i) x (subbytes res 0 i)))
-                      (else
-                       (bytevector-u8-set! res i x)
-                       (lp (+ i 1))))))))))
+        (read-bytevector! res in)
```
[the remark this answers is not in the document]
wasamasa: Good point, fixed

lib/scheme/extras.scm
```diff
@@ -135,36 +135,31 @@
   (if (zero? n)
       #u8()
       (let ((in (if (pair? o) (car o) (current-input-port)))
```
ashinn, line 137: `let*`

ashinn (approved):
Thanks!

lib/scheme/extras.scm
```diff
@@ -134,37 +134,32 @@
...
-                       (if (zero? i) x (subbytes res 0 i)))
-                      (else
-                       (bytevector-u8-set! res i x)
-                       (lp (+ i 1))))))))))
+      (let* ((in (if (pair? o) (car o) (current-input-port)))
+             (vec (make-bytevector n))
+             (res (read-bytevector! vec in)))
+        (cond ((eof-object? res) res)
+              ((< res n) (subbytes vec 0 i))
+              (else res)))))
```
ashinn, line 142:
```suggestion
              (else vec)))))
```

lib/scheme/extras.scm
```diff
@@ -134,37 +134,32 @@
...
-                (cond ((eof-object? x)
-                       (if (zero? i) x (subbytes res 0 i)))
-                      (else
-                       (bytevector-u8-set! res i x)
-                       (lp (+ i 1))))))))))
+      (let* ((in (if (pair? o) (car o) (current-input-port)))
+             (vec (make-bytevector n))
+             (res (read-bytevector! vec in)))
+        (cond ((eof-object? res) res)
+              ((< res n) (subbytes vec 0 i))
```
ashinn, line 141:
```suggestion
              ((< res n) (subbytes vec 0 res))
```

ashinn (changes requested):

Issue #953, opened 2024-05-15
Segmentation fault caused by GC with stale pointer on top of the VM stack


petteripiiroinen:
#### Brief description of the bug

Chibi VM can leave the stack top into an uninitialised and potentially stale 
state after `SEXP_OP_APPLY1` and `SEXP_OP_TAIL_CALL` instructions.
Under very particular circumstances this can trigger an out-of-bound error
and can cause a segmentation fault during marking phase in garbage collection.

* affected versions: Every Chibi Scheme commit starting at 2922ed591d1c0dc3be7a92e211ac7b18aa12edcc 
* platforms tested:
  * macOS 13.6.6, Intel x86-64 (Apple clang version 15.0.0 (clang-1500.1.0.2.5)
  * Debian Linux Bookworm, Intel x86-64 (gcc 12.2.0)
  * all platforms are affected

#### Way to reproduce the issue

1. prepare a test directory (I will use `${CHIBITEST}` for that and Bash as shell.)

```bash
mkdir ${CHIBITEST}
cd ${CHIBITEST}
```

2. clone the forked Chibi repository and checkout to `crash-demo-part1` branch.

Alternatively just create the following `crash-test.scm` file:

```scheme
(begin
  (define (spinning-loop n)
    (if (zero? n) #t
        (spinning-loop (- n 1))))
  (define (throwway-fill-heap n)
    (cond
      ((zero? n) #t)
      (else (make-vector 80000 0)
            (throwway-fill-heap (- n 1)))))
  (define (alternate-fill-heap-2 n keep? acc)
    (cond
      ((zero? n) acc)
      (keep? (alternate-fill-heap-2 (- n 1) #f (cons (cons 1 2) acc)))
      (else (alternate-fill-heap-2 (- n 1) #t acc))))
  (define (alternate-fill-heap n keep? acc)
    (cond
      ((zero? n) acc)
      (keep? (alternate-fill-heap (- n 1) #f (cons (cons 1 2) acc)))
      (else (alternate-fill-heap (- n 1) #t acc))))
  (define (stage-1)
    (write-string "crash test\n")
    (write-string "----------\n")
    (write-string "1. running VM for few million cycles for debugging:\n")
    (write-string "    - started \n")
    (spinning-loop 2360009)
    (write-string "    - done    \n"))
  (define (triggering-list n acc)
    (if (zero? n) acc
        (triggering-list (- n 1) (cons n acc))))
  (define (stage-2)
[...]

PR #954, opened 2024-05-15
Fix: segmentation fault during GC marking

petteripiiroinen:
The stack top needs to be initialised before a potential garbage collection after `SEXP_OP_APPLY1` and `SEXP_OP_TAIL_CALL`, since stack top can otherwise be pointing to a stale pointer.

This fixes the #953

### DIFF
--- a/vm.c
+++ b/vm.c
@@ -1095,7 +1095,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
   sexp_ensure_stack(i + 64 + (sexp_procedurep(tmp1) ? sexp_bytecode_max_depth(sexp_procedure_code(tmp1)) : 0));
   for (top += i; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
     _ARG1 = sexp_car(tmp2);
-  top += i+1;
+  top += i;
+  /* restore the make_call invariant */
+  _PUSH(tmp1);
   goto make_call;
 
  loop:
@@ -1266,7 +1268,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
       int prev_top = top;
       for (top=fp-j+i-1; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
         stack[top] = sexp_car(tmp2);
-      top = fp+i-j+1;
+      top = fp+i-j;
+      /* restore the make_call invariant */
+      _PUSH(tmp1);
       fp = k;
       /* if final cdr of tmp2 isn't null, then args list was improper */
       if (! sexp_nullp(tmp2)) {
@@ -1289,7 +1293,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
     /* copy new args into place */
     for (k=0; k<i; k++)
       stack[fp-j+k] = stack[top-1-i+k];
-    top = fp+i-j+1;
+    top = fp+i-j;
+    /* restore the make_call invariant */
+    _PUSH(tmp1);
     fp = sexp_unbox_fixnum(tmp2);
     goto make_call;
   case SEXP_OP_CALL:


vm.c
```diff
@@ -1299,6 +1299,14 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
   make_call:
     sexp_context_top(ctx) = top;
     if (sexp_opcodep(tmp1)) {
+      /* setting _ARG1 to SEXP_NULL prevents a segmentation fault caused
+       * by an unintialised stack top after SEXP_OP_APPLY1. It also
+       * prevents the same crash for SEXP_OP_TAIL_CALL when i
+       * - j >= 5 + m, where m is the number of used stack slots in the
+       * current frame
+       */
+      _ARG1 = SEXP_NULL;
```
ashinn, line 1308: Thank you for the detailed bug report, analysis, and fix!  I'm curious how you tracked this down.

It's still not clear to me if this is the right fix.  The stack only contains valid objects, and these will not be collected while still on the stack.

If we expand the stack (increase top) without filling in the value, that can result in a stale pointer on the stack.  But in that case we should fill in the value where we do the expansion.

Another possibility is how top and sexp_context_top differ.  The former is used as a register for quick stack access, but GC only knows about the latter.  If sexp_context_top is ever out of date when a GC triggers, that can cause the same problem, but then the fix is to sync the value before the GC.

ashinn (commented):
Thanks!

vm.c
petteripiiroinen: Hi, 

and thank you for your kind words and also thank you so much for creating Chibi. 

I have been (super slowly, over ten years, as a hobby project outside of my work) writing an r7rs (small) implementation called "Guovssat" (/ˈkuo̯vssah(t)/ Northern Sámi word, which in Finnish is "Kuukkeli" /ˈkuːkːeli/. The Latin name of this small bird is "Perisoreus infaustus" and in English this bird is known as the Siberian jay). I hope I will get the first version (once I have added the last major missing parts, namely full call/cc (I currently have a single shot call/cc), expose eval to users, stabilise the ABI of the libraries, rewrite the remaining C part (the bootloader) in Scheme together with a linker for (at least) ELF binaries so that the last C code parts could be dropped) out in a year (but most likely within two years). 

Guovssat Scheme is a bootstrapping non-optimising native code AOT incremental cross-compiler currently targeting 4 architectures (but I will add at least two more, but maybe after the first published version) and 4 host systems (macOS, Linux (both glibc and musl) and Windows 10) fully written in r7rs Scheme. 

I do call it a VM, even if it does not have a specific bytecode since the bytecode is native machine code on different architectures. I have been using Chibi both for quick check for r7rs conformancy validation and also as an secondary bootstrap option (i.e. running the Guovssat compiler under Chibi to compile Guovssat without it) to iron out the differences with r7rs. There were still few, since I started few years before r7rs small was ratified and never checked that the names and say how fold had changed.

Running a compiler under Chibi is stressing especially Chibi's memory management. Maybe I started getting very random crashes that just vanished when made any modifications in Sep 2022. Due to my super slow pace, and since I was rewriting part of my own GC subsystem, I could not always tell where the crash was coming.

I did not test Chibi as bootstrap option for a year or so since adding a full UTF-8 support, the Windows support, a support for Aarch64 (Arm-v8) and syntax-rules macros required quite a lot of changes and keeping Chibi changes were too much overhead for that. 

I need to modify the code for Chibi, since my implementation has some very low-level parts that are not in r7rs (so those need to be just dropped), and I wanted to move them to correct places namely under `(scheme ...)` libraries. While I was making it build again with Chibi I managed to get Chibi crash twice in a row. I tried to add a debug print and the crash vanished, but returned without the print.

With a reproducible crash, I just took a full copy of everything of the debug state that was crashing and I made sure it was still consistently crashing. Since I am benefiting from Chibi, I wanted to make a nice report on that and provide a good test for tracking the bug down.

I could not, however, produce a smaller test demonstrating the issue, so I had to track the actual reason for the crash, analyse the cause and try writing a small test demonstrating it. The original crash took typically 12-18 seconds to produce and with debug messages running around 40 seconds, so getting the short version that ran under 1 second was a nice bonus.

Since I have worked on multiple VM implementations and a GC implementation of my own and none of the ready-made debugging tools provide any real support for my use cases, I had developed ways to hunt down GC and other memory related issues. From the symptoms, I was sure that this is a GC issue, after all.

I usually need to find the exact place of the issue so I added global counters (the Finnish word "laskuri" /ˈlɑskuri/ means a counter, using Finnish helps making sure that I will not accidentally break something by using globals) and binary search to find the instruction VM was executing before crashing and this is how I found `SEXP_OP_APPLY1`.

Then I continued inside the marking (again with few other counters, since there are multiple loops there) to find the place and the time (in terms of the counters) of the crash (the line 268 in `gc.c`)

```c
static void sexp_mark_one (sexp ctx, sexp* types, sexp x) {
  ...
  t = types[sexp_pointer_tag(x)];
  ...
}
```

Then I gathered all the information there is: 
1. why it crashed (the tag value was `#xa0` which is outside of `types` array)
2. what was the path of getting there (a list (chain or pairs) of length `5`) inside a vector, inside of a closure and finally in stack (since it was coming from `sexp_context_top(ctx)`)
3. where in stack it was (just below the VM cached `top`)
4. where did that invalid tag `#xa0` come from (I added the relative pointers (of two types, since the test was so massive, there were four different types of cases out of which one of the types occured roughly 70 percent times, so I always just tested if we are in the case A and then placed the pointers I wanted to track into a global array with just hardcoded relative addresses for the case A for following when they are mutated). 
5. This is how I found that it was the sweeping that was producing the `#xa0` tag during previous sweeping on the line 514.

```c
sexp sexp_sweep (sexp ctx, size_t *sum_freed_ptr) {
  ...
  sexp p, end;
  sexp_free_list q, r, s;
  ...
  q->size = freed;
  ...
}
```

I needed to get some familiarity how Chibi is implemented, namely how allocation, GC and VM work, so then I played around of looking around how the VM loop works (how the push and pull are implemented, what is the frame layout, where are local variables in stack, in what order and where the calling arguments are placed on stack (related to `top`) and how `SEXP_OP_CALL` and `SEXP_OP_TAIL_CALL` are working (since all of these were going through `make_call:` label.

For that I wrote a small disassembling and the debug connecting facility for sending messages to Chibi during runtime, to find out what it is doing at what time. Once I understood enough how calls and tail calls were working, I could take a closer look at `SEXP_OP_APPLY1` and the real point (and the part where I fully agree with you, I don't really like the proposed fix myself either) for the issue are the lines 1267-1269 that copy the arguments from the list to the stack

```c
  case SEXP_OP_APPLY1:
    tmp1 = _ARG1;
    tmp2 = _ARG2;
  apply1:
    tmp = sexp_length(ctx, tmp2);
    ...
    i = sexp_unbox_fixnum(tmp); /* number of params */
    ...
    j = sexp_unbox_fixnum(stack[fp]);              /* previous num params */
    ...
    {
      int prev_top = top;
      for (top=fp-j+i-1; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
        stack[top] = sexp_car(tmp2);
      top = fp+i-j+1;
      ...
    }
    ...
    goto make_call;
```

Here, irregardless of the `fp` (frame pointer), irregardless of `i` and `j` (number of parameters passed to operator in `tmp1`, which is the length of list to be applied) and the number of caller parameters) we fill the stack starting from `N - 1` and going down by `i` to `N - i` where `N = fp - j + i`. I would have expected that `top` would then be set to `N` but it is set to `N + 1` leaving one stack slot uninitialised (and this is the place that brought the old pointer to become accidentally live).

I first just tested, could we change `top = fp + i - j` and that prevented Chibi to be built, so I had to think why this is done in this way. The easiest way to see it is to consider `SEXP_OP_CALL` that shares the `make_call` block of the code.

```c
  case SEXP_OP_CALL:
    _ALIGN_IP();
    i = sexp_unbox_fixnum(_WORD0);
    tmp1 = _ARG1;
  make_call:
```

In all the other cases (with two exceptions) the top of the stack is populated by the operator (the procedure object or the primitive opcode object). Thus, the joint `make_call` invariant is that at `make_call` time the stack is

| stack index | stack value       |
| ----------- | ----------------- |
| `N` = `top` | not in stack      |
| `N - 1`     | operator = `tmp1` |
| `N - 2`     | 1st argument      |
| ...         | ...               |
| `N - m + 1` | mth argument      |

The operator is cached into `tmp1` variable, so it is irrelevant, what is at the stack top (slot `N - 1`), but if we have a GC that time, it has to be a valid Chibi object.

This leads to the how to fix the issue and this is where there really are multiple ways, just like you mentioned.

I made the following suggestions (in order of preference, the prefered version first): 

1. suggested-fix-1 (74fac455) this corresponds to "If we expand the stack (increase top) without filling in the value, that can result in a stale pointer on the stack. But in that case we should fill in the value where we do the expansion."). This is safest and most correct. The cons are quite tiny, it only adds some extra CPU cycles during normal VM execution and typically in range of 100-1000 CPU cycles during GC.

```diff
diff --git a/vm.c b/vm.c
index e26f4ec5..7c032078 100644
--- a/vm.c
+++ b/vm.c
@@ -1095,7 +1095,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
   sexp_ensure_stack(i + 64 + (sexp_procedurep(tmp1) ? sexp_bytecode_max_depth(sexp_procedure_code(tmp1)) : 0));
   for (top += i; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
     _ARG1 = sexp_car(tmp2);
-  top += i+1;
+  top += i;
+  /* restore the make_call invariant */
+  _PUSH(tmp1);
   goto make_call;
 
  loop:
@@ -1266,7 +1268,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
       int prev_top = top;
       for (top=fp-j+i-1; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
         stack[top] = sexp_car(tmp2);
-      top = fp+i-j+1;
+      top = fp+i-j;
+      /* restore the make_call invariant */
+      _PUSH(tmp1);
       fp = k;
       /* if final cdr of tmp2 isn't null, then args list was improper */
       if (! sexp_nullp(tmp2)) {
@@ -1289,7 +1293,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
     /* copy new args into place */
     for (k=0; k<i; k++)
       stack[fp-j+k] = stack[top-1-i+k];
-    top = fp+i-j+1;
+    top = fp+i-j;
+    /* restore the make_call invariant */
+    _PUSH(tmp1);
     fp = sexp_unbox_fixnum(tmp2);
     goto make_call;
   case SEXP_OP_CALL:
```

2.  suggested-fix-2 (abdf3630) this still corresponds to "If we expand the stack (increase top) without filling in the value, that can result in a stale pointer on the stack. But in that case we should fill in the value where we do the expansion."). This is faster than the previous, but I can imagine a scenario where the Opcode might get corrupted during unfortunate GC (if the Opcode is the last live version) and that might cause havoc during the bytecode compilation.

```diff
diff --git a/vm.c b/vm.c
index e26f4ec5..71c858d1 100644
--- a/vm.c
+++ b/vm.c
@@ -1095,7 +1095,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
   sexp_ensure_stack(i + 64 + (sexp_procedurep(tmp1) ? sexp_bytecode_max_depth(sexp_procedure_code(tmp1)) : 0));
   for (top += i; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
     _ARG1 = sexp_car(tmp2);
-  top += i+1;
+  top += i;
+  /* initialise stack top for make_call */
+  _PUSH(SEXP_NULL);
   goto make_call;
 
  loop:
@@ -1266,7 +1268,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
       int prev_top = top;
       for (top=fp-j+i-1; sexp_pairp(tmp2); tmp2=sexp_cdr(tmp2), top--)
         stack[top] = sexp_car(tmp2);
-      top = fp+i-j+1;
+      top = fp+i-j;
+      /* initialise stack top for make_call */
+      _PUSH(SEXP_NULL);
       fp = k;
       /* if final cdr of tmp2 isn't null, then args list was improper */
       if (! sexp_nullp(tmp2)) {
@@ -1289,7 +1293,9 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
     /* copy new args into place */
     for (k=0; k<i; k++)
       stack[fp-j+k] = stack[top-1-i+k];
-    top = fp+i-j+1;
+    top = fp+i-j;
+    /* initialise stack top for make_call */
+    _PUSH(SEXP_NULL);
     fp = sexp_unbox_fixnum(tmp2);
     goto make_call;
   case SEXP_OP_CALL:
(END)

```

3. suggested-fix-3 (a8ad7541) this corresponds to "Another possibility is how top and sexp_context_top differ. ... The former is used as a register for quick stack access, but GC only knows about the latter"). I don't like this, it really feels like a hack around the issue. Trying to localise the having a non-sync state was even more hacky, so I don't even write that. These is also even more unsafe than the previous.

```diff
diff --git a/vm.c b/vm.c
index e26f4ec5..dccf2d57 100644
--- a/vm.c
+++ b/vm.c
@@ -1297,15 +1297,19 @@ sexp sexp_apply (sexp ctx, sexp proc, sexp args) {
     i = sexp_unbox_fixnum(_WORD0);
     tmp1 = _ARG1;
   make_call:
-    sexp_context_top(ctx) = top;
     if (sexp_opcodep(tmp1)) {
+      /* don't expose the potentially uninitialise stack top */
+      sexp_context_top(ctx) = top - 1;
       /* compile non-inlined opcode applications on the fly */
       tmp1 = make_opcode_procedure(ctx, tmp1, i, SEXP_PROC_NONE);
       if (sexp_exceptionp(tmp1)) {
+        /* resync the stack top in context */
+        sexp_context_top(ctx) = top;
         _ARG1 = tmp1;
         goto call_error_handler;
       }
     }
+    sexp_context_top(ctx) = top;
     if (! sexp_procedurep(tmp1))
       sexp_raise("non procedure application", sexp_list1(ctx, tmp1));
     j = i - sexp_procedure_num_args(tmp1);
```

Both of these assume that on every occasion of having `Opcode` object as the operator this object has a live and valid pointer somewhere (either somewhere in stack or in a global root system). This surely is valid for `SEXP_OP_CALL` since it was pushed somewhere to the stack and that place still is live. I also noticed that the `sexp_apply` can trigger this, but I think only when Chibi is used embedded in an application.

Thus: I propose that the current proposed fix is replaced with the first of the candidates. If it is ok, I would rebase this on top of the current Chibi head, and change the original proposal with the new one (by force pushing to update the pull request). If you have even a better alternative in mind, I would like to hear that. I would prefer not to have two commits for the fix, one original and then correcting commit.

After reading the `SEXP_OP_CALLCC` case it is not effected with this since it starts with

| stack index | stack value       |
| ----------- | ----------------- |
| `N + 4`     | not in stack      |
| `N + 3`     | not in stack      |
| `N + 2`     | not in stack      |
| `N + 1`     | not in stack      |
| `N ` = `top`| not in stack      |
| `N - 1`     | 1st argument      |
| ...         | ...               |
| `N - m + 1` | mth argument      |

and at `make_call` it is

| stack index      | stack value       |
| ---------------- | ----------------- |
| `N + 4`          |                   |
| `N + 3`          | frame info        |
| `N + 2`          | frame info        |
| `N + 1` = `top`  | frame info        |
| `N `             | SEXP_ONE          |
| `N - 1`          | procedure         |
| ...              | ...               |
| `N - m + 1`      | mth argument      |

so `SEXP_OP_CALLCC` cannot trigger this issue.

ashinn: Wow, that's a lot of investigative work!  Thank you very much for the detail.  Looking forward to seeing Guovssat.

I agree the new suggested fix looks best, will apply.

ashinn (approved):
Thanks!

PR #956, opened 2024-05-21
Added missing argument to `sexp_string_cursor_set`

0xJonas:
The macro was missing the `ch` argument, causing calls to that macro to not compile.


ashinn (approved):
Thanks for catching this!

PR #980, opened 2024-05-28
Add and test array-fold-{left|right}, remove array-fold{l|r}

gambiteer:
The commit history is a bit ragged, sorry.


gambiteer:
I have to rethink how this works, I apologize for the noise, your tests don't pass.

ashinn:
The tests are from SRFI 179.  They fail because they're fold-left on matrix multiplication which is not commutative.  Just updating the tests is fine.

PR #990, opened 2024-06-02
Fix SEXP_USE_ALIGNED_BYTECODE

welcome-linja:
There were some typos that weren't affecting default compilation but causing immediate errors with the option enabled.

### DIFF
--- a/include/chibi/features.h
+++ b/include/chibi/features.h
@@ -301,7 +301,7 @@
 
 /* uncomment this to make the VM adhere to alignment rules */
 /*   This is required on some platforms, e.g. ARM */
-/* #define SEXP_USE_ALIGNED_BYTECODE */
+/* #define SEXP_USE_ALIGNED_BYTECODE 1 */
 
 /************************************************************************/
 /* These settings are configurable but only recommended for */


include/chibi/features.h
```diff
@@ -301,7 +301,7 @@
 
 /* uncomment this to make the VM adhere to alignment rules */
 /*   This is required on some platforms, e.g. ARM */
-/* #define SEXP_USE_ALIGNED_BYTECODE */
+/* #define SEXP_USE_ALIGNED_BYTECODE 1 */
```
ashinn, line 304: Note these two lines are equivalent, the 1 is implied by the preprocessor.

include/chibi/sexp.h
```diff
@@ -1445,7 +1445,7 @@ enum sexp_uniform_vector_type {
 #define sexp_context_max_depth(x) (sexp_vector_ref(sexp_context_specific(x), SEXP_FIVE))
 #define sexp_context_exception(x) (sexp_vector_ref(sexp_context_specific(x), SEXP_SIX))
 
-#if SEXP_USE_ALIGNED_BYTECODE
+#ifdef SEXP_USE_ALIGNED_BYTECODE
```
ashinn, line 1448: This is only required because you removed the line above where it's set to 0.

ashinn (commented):
I don't understand what you're fixing here.  You're just making this macro inconsistent with all other features by making it defined/undefined instead of 1/0.

include/chibi/sexp.h
welcome-linja: Actually no, that line was erroring compilation even when all I'd done was uncomment the `#define`. I don't know if I have a more strict compiler than you do but for me it was either this or `#if SEXP_USE_ALIGNED_BYTECODE == 1`.

welcome-linja:
> I don't understand what you're fixing here. You're just making this macro inconsistent with all other features by making it defined/undefined instead of 1/0.

What I'm fixing is previously, just uncommenting the option would cause errors:
``` sh
echo '#define sexp_so_extension "'.dylib'"' > include/chibi/install.h
echo '#define sexp_default_module_path "'/usr/local/share/chibi:/usr/local/lib/chibi:/usr/local/share/snow:/usr/local/lib/snow'"' >> include/chibi/install.h
echo '#define sexp_platform "'macosx'"' >> include/chibi/install.h
echo '#define sexp_architecture "'arm64'"' >> include/chibi/install.h
echo '#define sexp_version "'0.11.0'"' >> include/chibi/install.h
echo '#define sexp_release_name "'`cat RELEASE`'"' >> include/chibi/install.h
cc -c  -Iinclude  -DSEXP_USE_NTPGETTIME -DSEXP_USE_INTTYPES -Wall -g -g3 -O3  -o main.o main.c
In file included from main.c:9:
In file included from include/chibi/eval.h:12:
include/chibi/sexp.h:1448:30: error: expected value in expression
#if SEXP_USE_ALIGNED_BYTECODE
                             ^
1 error generated.
make: *** [main.o] Error 1
```

I can make it 1/0 if that's the style, but the current way is broken.

welcome-linja:
There you go.

include/chibi/features.h
ashinn: Nevermind, I'm thinking of the -D command-line option.

ashinn:
Sorry, I thought you were reporting a bug in the code, this is now just a documentation fix and looks OK.

PR #993, opened 2024-06-16
README: add link to online manual

il-k:
It is not uncommon that the git repo is the first encounter with a project. Having the manual available in the repo makes it easier to discover.


ashinn (approved):
Thanks!

PR #998, opened 2024-08-25
(chibi test): add a type test for exceptions in test-error

dpk:
Arguably the `name` and `pred` should be independent of one another, but that would require rewriting `test-error` as an explicit renaming macro in order to be able to type check the `name` (and assume it must be a string) and thus lose portability.


ashinn (approved):
Thanks!

Issue #883, opened 2023-01-29
Square brackets


mnieper:
Would you accept a patch that enables using square brackets (as used in R6RS, for example) to improve portability?

lassik:
How different is `#!r6rs` lexical syntax from R7RS-small syntax? If not much, the easiest way to do it might be to recognize `#!r6rs`.

mnieper:
> How different is `#!r6rs` lexical syntax from R7RS-small syntax? If not much, the easiest way to do it might be to recognize `#!r6rs`.

This is somewhat unrelated to the question.  The usual meaning of `#!r6rs` is to force signaling errors for all lexical syntax that is not R6RS.  This is only sensible for an R6RS implementation, not for Chibi.

The context of the question is different.  R7RS is compatible with the lexical syntax extension I asked about.

ashinn:
The meaning of square brackets should be configurable.  I think a reasonable approach is to allow specifying the identifier to which they expand similar to Kawa:

```
#|kawa:1|# '[foo]
($bracket-list$ foo)
```

Here you can locally override the binding of `$bracket-list$`.  This is insufficient for R6RS.  As a special case, if the identifier is `#f`, then the expansion doesn't include the identifier.

mnieper:
So, `[...]`  would be a similar abbreviation as quote in the reader. (Only that `quote` is non-configurable.)

As it is a configuration of the reader, I wonder where the identifier (or ` #f`) has to be set to make it as helpful as possible.  It could be a command-line option for Chibi, which is better than a compile-time flag.

ashinn:
Yes, runtime is preferable.  We can consider:

1. global setting (configured via command-line option and/or code)
2. per-extension setting (e.g. by default .sls would follow R6RS and .scm would follow Kawa)
3. per-file setting (via `#!` pragmas and/or allowing the library declaration language to customize for `include` files)

1 is simplest for now so I recommend that unless you feel ambitious.

Beware: there are _two_ implementations of `read` in Chibi.

lassik:
IMHO read and write syntax settings should be bound to port objects. I did some sketching of this a while back but didn't manage to come up with an elegant API.

Gambit also supports something like Kawa's approach IIRC. You should ask Feeley about that. I don't like the approach, but he does.

mnieper:
> IMHO read and write syntax settings should be bound to port objects. I did some sketching of this a while back but didn't manage to come up with an elegant API.
> 
[...]

PR #1003, opened 2024-10-06
Partial R6RS library compatibility

dpk:
Aiming for Guile/Larceny-style R6RS compatibility, rather than Chez-style R6RS total conformance. A correct R6RS library will work, but error cases will reflect Chibi’s existing style rather than what R6RS strictly demands.

Still to do:

- [ ] I’m not sure `clean-up-r6rs-import` handles all cases yet
- [ ] Support `.sls` extension for library files, looking for `.chibi.sls` first
- [ ] Decide what to do about SRFI 97-style library names (add redundant libraries for all of them, as in Arcfide’s chez-srfi project?)
- [ ] Decide what to do about square brackets
- [ ] Add the remaining `(rnrs ...)` libraries, probably piggybacking off the R6RS-cleanup SRFIs like 181 and 237/240 where appropriate
  - [x] Decide how to support R6RS conditions within Chibi’s R7RS error object style system
- [ ] Modify the expander to use the R6RS order at least for procedure and library bodies (which is R7RS-compatible; there is a deeper issue with program bodies, see <https://github.com/larcenists/larceny/issues/825>)

If there is interest I will do these things. The only ‘core’ change is the one already made in `meta-7.scm`; for total support it also needs the macro expander body processing order change, which I don’t expect to be a massive change either. For square brackets it depends on the overall approach taken (#883).

### DIFF
--- a/lib/rnrs/base.sld
+++ b/lib/rnrs/base.sld
@@ -190,12 +190,16 @@
           vector-set!
           vector?
           zero?)
-  (import (rename (scheme base)
-                  (error r7rs:error))
+  (import (except (scheme base)
+                  define-syntax
+                  let-syntax
+                  letrec-syntax
+                  syntax-rules)
           (scheme cxr)
           (scheme inexact)
           (scheme complex)
           (rnrs conditions)
+          (only (srfi 1) every)
           (rename (srfi 141)
                   (euclidean-quotient div)
                   (euclidean-remainder mod)
@@ -203,8 +207,37 @@
                   (balanced-quotient div0)
                   (balanced-remainder mod0)
                   (balanced/ div0-and-mod0))
+          (rename (chibi syntax-case)
+                  (splicing-let-syntax let-syntax)
+                  (splicing-letrec-syntax letrec-syntax))
           (except (chibi ast) error)
           (chibi show))
+
+  (define-syntax syntax-rules
+    (lambda (x)
+      (syntax-case x ()
+        ((_ (lit ...) ((k . p) t) ...)
+         (every identifier? #'(lit ... k ...))
+         #'(lambda (x)
+             (syntax-case x (lit ...)
+               ((_ . p) #'t) ...))))))
+
+  (define-syntax identifier-syntax
+    (lambda (x)
+      (syntax-case x (set!)
+        ((_ e)
+         #'(lambda (x)
+             (syntax-case x ()
+               (id (identifier? #'id) #'e)
+               ((_ x (... ...)) #'(e x (... ...))))))
+        ((_ (id exp1) ((set! var val) exp2))
+         (and (identifier? #'id) (identifier? #'var))
+         #'(make-variable-transformer
+            (lambda (x)
+              (syntax-case x (set!)
+                ((set! var val) #'exp2)
+                ((id x (... ...)) #'(exp1 x (... ...)))
+                (id (identifier? #'id) #'exp1))))))))
 
   (define-syntax assert
     (syntax-rules ()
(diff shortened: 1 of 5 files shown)


lib/meta-7.scm
```diff
@@ -355,6 +355,71 @@
...
+       (eq? (strip-syntactic-closures id_1)
+            (strip-syntactic-closures id_2)))
+     (define (clean-up-r6rs-library-name name)
+       (define (srfi-number->exact-integer component)
+         (if (symbol? component)
+             (let* ((symbol-name (symbol->string component))
+                    (maybe-number-as-string (substring symbol-name 1)))
+               (if (and (char=? (string-ref symbol-name 0) #\:)
+                        (every char-numeric?
+                               (string->list maybe-number-as-string)))
```
ashinn, line 371: nit: You can get rid of the substring:

```suggestion
                               (cdr (string->list symbol-name))))
```

lib/meta-7.scm
```diff
@@ -355,6 +355,71 @@
...
+     (let ((library-name (clean-up-r6rs-library-name (list-ref expr 1)))
+           (exports (cdr (list-ref expr 2)))
+           (imports (map clean-up-r6rs-import (cdr (list-ref expr 3))))
+           (body (cddr (cddr expr)))
+
+           (_define-library (rename 'define-library))
+           (_export (rename 'export))
+           (_import (rename 'import))
+           (_begin (rename 'begin)))
+       `(,_define-library ,library-name
```
ashinn, line 416: Both the R6RS and R7RS library languages are static and outside the Scheme language itself, and can't in general be expanded into by macros.  So you shouldn't need to compare with `symbolic-id=?` here.

On the other hand, the existing `define-library` _also_ doesn't expect macro expansions, and compares symbols directly with `eq?`, so I think you either need to use the raw symbols here (no renaming) or use `identifier=?` in `define-library`.

lib/rnrs/base.sld
```diff
@@ -0,0 +1,231 @@
...
+                  (euclidean-quotient div)
+                  (euclidean-remainder mod)
+                  (euclidean/ div-and-mod)
+                  (balanced-quotient div0)
+                  (balanced-remainder mod0)
+                  (balanced/ div0-and-mod0))
+          (except (chibi ast) error)
+          (chibi show))
+
+  (define-syntax assert
```
ashinn, line 208: Is `(chibi assert)` a proper superset of R6RS assert?  It gives better reporting.

ashinn (commented):
Thanks!  In general I'm OK with a change of this size for now, but the key question is how to handle the file extensions.  It was always the intention to support other languages (notably javascript), and load a different parser & evaluator based on the file extension, and that should be handled dynamically.

But if we can do so with minimal core changes I'm OK with more native R6RS support.

lib/meta-7.scm
ashinn: `define-library` doesn't consistently strip syntactic closures so I'm not sure this will always work.  It's safer to leave it as a symbol.

ashinn:
> The other ‘core’ changes, as noted above, are: square bracket support in the reader (#883);

Support for customizing square brackets is fine.  Note it's likely the default behavior will become square brackets as array literals for compatibility with Remora.

> supporting `.sls` as a file extension for libraries (even this is not strictly necessary since R6RS doesn’t say anything about file names, but `.sls` is the convention in the same way `.sld` is on R7RS)

Probably not an issue but there's the question of what to do when both `.sld` and `.sls` are found, and the general concern about the slowdown when searching for more files.

> and changing the expander to expand all syntax definitions before expanding all other definitions and expressions in procedures and libraries.

This may be a problem.

PR #1008, opened 2024-10-26
Add SRFI 35 support

dpk:
This PR is part of the work for #1003.

It adds support for [SRFI 35](https://srfi.schemers.org/srfi-35/srfi-35.html), with an internal module including extra exports necessary to implement the `(rnrs conditions)` library.

It also adds support for printing conditions to the `(chibi repl)` module.

It also fixes a bug in the SRFI 99 implementation whereby shadowing field names in the procedural layer didn’t work. There is a bug in the design of SRFI 99 which means shadowing still doesn’t work if you pass `rtd-constructor` an explicit list of field names, but this at least fixes the problem for uses which don’t do this.


ashinn:
> I should note there is still a problem with the REPL: it should return from continuable non-serious conditions (since in R6RS conditions are also used for warnings). At the moment it assumes all exceptions raised should not be continued. But this is a somewhat larger rewiring of the REPL’s exception-handling mechanism, so I haven’t attempted it.

I'm not sure what you mean here.  The REPL puts no restrictions on exception handling, however if an exception is _uncaught_ and escapes to the top-level the REPL is free to do what it wants, including crashing.  For user convenience we instead print the exception and provide a new prompt.  If you mean you want an interactive debugger with the ability to _manually_ continue an exception, then that's a much larger project and clearly outside the scope of standards.

lib/chibi/repl.sld
```diff
@@ -8,6 +8,7 @@
           (srfi 1)
           (srfi 9)
           (only (srfi 18) current-thread)
+          (srfi 35 internal)
```
ashinn, line 11: My understanding was that R7RS Large was actually going in a different direction from R6RS conditions?

R7RS small uses opaque predicates, and John wanted a list of condition-types: https://github.com/johnwcowan/r7rs-work/blob/master/ConditionsCowan.md

I don't want to pull anything into the core which is at odds with R7RS.  I also don't want to bloat the core.  Many libraries are going to want to provide nicer pretty printing for the REPL (I want this for arrays).  To manage this we should either use a generic which can be overridden per type, or provide hooks which would be called by the other libraries (i.e. on loading SRFI 35 it would register itself into the REPL).

lib/srfi/35/internal.scm
```diff
@@ -0,0 +1,230 @@
...
+           (proc c))
+          ((find (lambda (comp) (condition-has-type? comp ct))
+                 (compound-condition-components c))
+           => (lambda (comp)
+                (proc comp)))
+          (else (error "condition-accessor: condition does not have the right type"
+                       c ct)))))
+
+(define-syntax define-condition-type/constructor
+  (er-macro-transformer
```
ashinn, line 132: I guess syntax-rules is easier here?

ashinn (commented):
Thanks!

ashinn:
> If the raised exception is a SRFI 35 condition, and does not have a condition type `&serious` (i.e. not `serious-condition?`), the REPL’s default exception handler should return some unspecified value, instead of calling the continuation it captures to escape back to the prompt.

... but there is no REPL in R6RS, isn't the behavior unspecified?

lib/srfi/35/internal.scm
```diff
@@ -0,0 +1,206 @@
+(define-record-type Simple-Condition
+  (make-simple-condition)
+  simple-condition?)
+
+(define-record-type Compound-Condition
```
ashinn, line 5: On reviewing the SRFI I'm rather unhappy with it because of this distinction.  It's ad-hoc and leads to a lot of ambiguity which doesn't exist in other exception systems.  Cleaner is to always have one type which may have fields referring to other types (e.g. the underlying causes).  I guess this was meant to make up for the lack of multiple inheritance.

lib/srfi/35/internal.scm
```diff
@@ -0,0 +1,206 @@
+(define-record-type Simple-Condition
```
ashinn, line 1: I have to think about this, but long term it may be better to unify the native exceptions with Simple-Condition.

lib/chibi/repl.scm
```diff
@@ -410,6 +453,8 @@
      (lambda (n) (thread-interrupt! thread))
      (lambda ()
        (protect (exn
+                 ((condition? exn)
```
ashinn, line 456: Please remove this.  I just pushed a change allowing customization via the new `repl-print` generic.  If you want, loading `(srfi 35)` can automate this customization with:

```
(define-method (repl-print (c condition?) (out output-port?))
  (repl-print-condition c out))
```

or otherwise provide a convenience procedure to register the method.

ashinn (commented):
Thanks!

lib/chibi/repl.scm
ashinn: Oops, added repl-print-exception as well.

lib/srfi/35/internal.scm
ashinn: Sorry, I wasn't clear in my complaint.  The complaint is in the existence of, and design of, compound conditions, not any implementation details.

The fundamental problem is that in SRFI 35 a given condition doesn't have any single, or even most specific, type.  You move the burden of taking apart and understanding the condition to the call site, which isn't in general prepared or qualified to deal with this.  Instead the code raising the condition should take care to choose a single most appropriate type.  If there is a root cause it can be attached as a field, so that where it really matters the caller can unwrap and understand the condition in more detail.

ashinn (approved):
Thanks!  LGTM,

PR #1017, opened 2025-01-18
Error on unfulfilled cond-expand.

a2379:
This complies with the SRFI 0 specification, and avoids the undefined behavior of R7RS.


ashinn:
Thanks!  I have to think about this though.  R7RS leaves this unspecified, so there's no requirement to signal an error, and there may be many existing programs relying on the fact that Chibi doesn't throw an exception.

PR #1024, opened 2025-06-08
add snow-chibi default configuration file instructions

Pinjontall94:
When running snow-chibi it errored out saying it needed a configuration file at `~/.snow/config.scm`, but it took a little trial and error to figure out both the minimum and my packaging use cases. Figured it might help others use it in the future!

### DIFF
--- a/doc/chibi.scrbl
+++ b/doc/chibi.scrbl
@@ -1446,17 +1446,11 @@ with image files on your platform you can run
 
 \subsection{Snow Configuration}
 
-Snow is configured in the file $HOME/.snow/config.scm, requiring at the
-bare minimum an empty alist:
-\schemeblock{
-()
-;; (()) would also work
-}
-
-Note that the empty list is \emph{not} quoted! This file is not evaluated,
-but is simply read as an alist of configurations. A more common example,
-for use with packaging your own code, may look like this:
-
+Snow is configured in the file $HOME/.snow/config.scm. A common example,
+for use with packaging your own code, may look like the one below:
+(Note that the empty list is \emph{not} quoted! This file is not evaluated,
+but is simply read as an alist of configurations. See
+https://snow-fort.org/doc/author/ for more details.)
 \schemeblock{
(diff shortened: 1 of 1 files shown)


doc/chibi.scrbl
```diff
@@ -1444,6 +1444,26 @@ loading (since it loads many libraries) - if you have any difficulties
 with image files on your platform you can run
 \command{snow-chibi --noimage} to disable this feature.
 
+\subsection{Snow Configuration}
+
+Snow is configured in the file $HOME/.snow/config.scm, requiring at the
+bare minimum an empty alist:
+\schemeblock{
+()
+;; (()) would also work
```
ashinn, line 1453: Not sure this is safe?

ashinn (approved):
Thanks!

doc/chibi.scrbl
ashinn: It's just a matter of whether the code always reliably checks for a proper alist.  Even if it works I wouldn't advertise this.

This config is the spec from `snow-chibi.scm`, there should really be an option to verify it automatically (using `conf-verify`), but it's a little non-trivial because it supports multiple subcommand configs as described in `app.scm`.

Anyway, just a basic example is sorely lacking, thanks for adding this.  For the record, my own config is:

```
((sign-uploads? #f)
 (use-curl? #t)
 (command
  (package
   (authors "Alex Shinn <alexshinn@gmail.com>")
   (maintainers "Alex Shinn <alexshinn@gmail.com>")
   (doc-from-scribble #t)
   (version-file "VERSION")
   (test-library (append-to-last -test))
   (license bsd))))
```

Pinjontall94: Nice! I'll just leave up the basic author, maintainers, and license config then :)

ashinn: sgtm

Pinjontall94:
How about this?

ashinn (approved):
LGTM

PR #1025, opened 2025-06-14
Add bash completion for the snow-chibi command names

Retropikzel:
Very basic bash completion for the command names. I will propably add more after I learn to use all the commands better.


Retropikzel:
Thank you.

PR #1026, opened 2025-06-15
Add generic implementation into snow-chibi

Retropikzel:
These changes add generic implementation to install libraries to be used with unsupported implementations.


lib/chibi/snow/utils.scm
```diff
@@ -25,6 +25,10 @@
             ,(delay
                (process->sexp
                 '(foment -e "(write (features))"))))
+    (generic "generic" #f #f
+            ,(delay
+               (process->sexp
+                '(echo "generic"))))
```
ashinn, line 31: I guess we can just `(write-string "generic\n")` instead of using a sub-process.

ashinn (changes requested):
Thanks!

Retropikzel:
I noticed that loko scheme creates path /usr/local/share/r6rs. Any thoughts on should the generic implementation use similar paths? Like /usr/local/lib/r7rs or /usr/local/share/r7rs?

ashinn:
> I noticed that loko scheme creates path /usr/local/share/r6rs. Any thoughts on should the generic implementation use similar paths? Like /usr/local/lib/r7rs or /usr/local/share/r7rs?

It's presumptuous if there are competing R7RS package managers.  I think it's safest to use a snow-specific path for now.  People can symlink if need be.

ashinn (approved):
LGTM

PR #1027, opened 2025-06-16
Fix propably misplaced parenthesis on get-install-dirs with chicken

Retropikzel:
If chicken is not installed then make-path on this line complains that #f is not valid input. I'm assuming this is what the code originally intended to do? Use install-prefix if it set, otherwise use "lib".


ashinn (approved):
Thanks for the fix!

Retropikzel:
Thank you.

PR #1028, opened 2025-06-16
Add stklos support for snow-chibi

Retropikzel:
Note that in version 2.10 stklos does not have the recommended behaviour for include.
It is however fixed in git. https://github.com/egallesio/STklos/issues/694


ashinn (approved):
LGTM

Retropikzel:
Thank you.

PR #1029, opened 2025-06-20
snow-chibi --foreign-depends

Retropikzel:
### Add support for --foreign-depends

I have a library that depends on libffi and I could not find how to add (foreign-depends "ffi") into the package.scm. It is read on package installation and library flags are passed to the compiler. So the support on that side exists. I added a --foreign-depends option to add it into package.scm during packaging. Though there is a change that I just missed how it is supposed to be done.

### Fix possible library install compilation bug

While packaging and installing my library I think I run into a bug. In commands.scm chibi-builder procedure if .stub file exists then chibi-ffi is run with options to also try to compile the library. This puts the -lffi flag on compilation command not at the end with -lchibi-scheme but at the beginning and the resulted .so file can not be loaded due to missing linkin to libffi.

Rather than begin to fix the chibi-ffi I switched the order of checking. The stub file is now checked first and if it exists but the C file does not then chibi-ffi is used to only generate C file and then C file check and compilation command are run.

I have not checked but perhaps the usual way to share ffi code has been with C files, not .stub files and because of thas no one noticed?

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -2051,22 +2051,30 @@
                  (lib-flags
                   (map (lambda (lib) (string-append "-l" lib))
                        (library-foreign-dependencies impl cfg library)))
-                 (ffi-cmd `(,@chibi-ffi ,stub-file))
+                 (ffi-cmd
+                  `(,@chibi-ffi
+                    "-c" "-cc" ,(car cc)
+                    "-f" ,(string-join cflags " ")
+                    "-f" ,(string-join lib-flags " ")
+                    ,@(if local-test? '("-f" "-Iinclude -L.") '())
+                    ,@(if (pair? (cdr cc))
+                          (list "-f" (string-join (cdr cc) " "))
+                          '())
+                    ,stub-file))
                  (cc-cmd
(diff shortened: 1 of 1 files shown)


Retropikzel:
I reverted the changes made to the ffi and compile commands. This pull request is now purely about the --foreign-depends flag.

doc/chibi.scrbl
```diff
@@ -1613,6 +1613,11 @@ can specify any option, for example:
    (license gpl))))
 }
 
+\itemlist[
+\item{\scheme{--foreign-depends} - specify foreign libraries the library
```
ashinn, line 1617: Thanks, this is fine for now informationally.  Ideally we'd verify it automatically before installation.

Also it can be inferred in some cases, e.g. if the impl is Chibi then an include-shared containing a file with a `(link "sqlite3")` declaration makes the dependency explicit.

ashinn (approved):
LGTM

doc/chibi.scrbl
Retropikzel: I need this for Gauche, Chicken and Gambit(in the future) too. Where would it be checked from automatically?

PR #1031, opened 2025-06-27
snow-chibi Kawa improvements

Retropikzel:
Adds .class file compilation to snow-chibi for libraries when installing for Kawa.

Since /usr/local/share/kawa looks like this:
```
/usr/local/share/kawa/
├── bin
│   └── kawa
└── lib
    └── kawa.jar
```
I have also changed the installation path from /usr/local/share/kawa to /usr/local/share/kawa/lib. I will try to get changes merged into Kawa, so that this new path would be added by default when running Kawa with the kawa script. I do not think they would add the previous one.


Retropikzel:
Thank you!

PR #1032, opened 2025-06-28
Add support for Mosh to snow-chibi

Retropikzel:
No eval so little bit ugly tmpfile hack used to get library path.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -1401,8 +1401,8 @@
     ((mosh)
      (with-output-to-file
       (make-path (cond-expand (windows (get-environment-variable "TMP"))
-                                  (else "/tmp"))
-                      "snowmosh")
+                              (else "/tmp"))
+                 "snowmosh")
       (lambda ()
        (display "(import (scheme base) (scheme write) (mosh config))")
        (newline)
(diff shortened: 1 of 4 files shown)


lib/chibi/snow/commands.scm
```diff
@@ -1398,6 +1398,16 @@
...
+    ((mosh)
+     (with-output-to-file
+      (make-path (cond-expand (windows (get-environment-variable "TMP"))
+                              (else "/tmp"))
+                 "snowmosh")
+      (lambda ()
+       (display "(import (scheme base) (scheme write) (mosh config))")
+       (newline)
+       (display "(display (get-config \"library-path\"))")))
+     (list (make-path (process->sexp '(mosh /tmp/snowmosh)) "lib")))
```
ashinn, line 1410: This won't work on Windows if TMP is not /tmp.

lib/chibi/snow/commands.scm
```diff
@@ -1398,6 +1398,16 @@
         (if (string? path)
             path
             "/usr/local/share/guile/"))))
+    ((mosh)
+     (with-output-to-file
+      (make-path (cond-expand (windows (get-environment-variable "TMP"))
```
ashinn, line 1403: Will this always be set?  Is there a safe default, or should we just use "."?

ashinn (changes requested):
Thanks!

Retropikzel:
I will fix this after my other pull request is merged and from then on will open only one pull request at a time.

Retropikzel:
Thank you!

PR #1033, opened 2025-06-28
Add support for Racket to snow-chibi

Retropikzel:
If you want to test run:
`raco pkg install --auto r7rs`
and then the script with:
`racket -I r7rs --script FILE.scm`


Retropikzel:
Thank you!

PR #1038, opened 2025-07-15
Add support for Ypsilon in snow-chibi

Retropikzel:


lib/chibi/snow/utils.scm
```diff
@@ -67,7 +67,18 @@
...
                     (process->sexp
-                     '(stklos -e "(write (features))"))))))
+                     '(stklos -e "(write (features))"))))
+    (ypsilon "ypsilon" (ypsilon --version) #f
+             ,(delay
+                (call-with-temp-file "snow-ypsilon"
+                 (lambda (tmp-path out preserve)
+                  (with-output-to-file tmp-path
+                                       (lambda ()
+                                        (display "(import (scheme base) (scheme write) (mosh config))")
```
ashinn, line 77: Is `(mosh config)` right here?

Retropikzel: No. Forgot to remove it when copying from Mosh support code. I will remove it.

PR #1040, opened 2025-07-29
Build static library too when installing for Chicken

Retropikzel:
Statically built library files are installed in addition so shared libraries when installing for Chicken scheme.


Retropikzel:
Thank you.

PR #1042, opened 2025-09-02
Allow package install trough git repository

Retropikzel:
Based on your comments here https://groups.google.com/g/chibi-scheme/c/MAVZqNo42Ps I started work on adding git support to snow-chibi.

The package in repo.scm looks like this:

```
 (package     (git (url "https://git.sr.ht/~retropikzel/scheme-hello-world")
                          (tag "v1.2.1")
                          (hash "1778bb98469e33b70ac1cde46524961007f847a7"))
                     (maintainers "Retropikzel <retropikzel@iki.fi>")
                     (authors "Retropikzel <retropikzel@iki.fi>")
                     (version "1.2.1")
                     (library (name (retropikzel hello)) (path "retropikzel/hello.sld") (foreing-depends) (depends (scheme base) (scheme write)))
                     (library (name (retropikzel hello util)) (path "retropikzel/hello/util.sld") (foreign-depends) (depends (scheme base) (scheme write)))
                     (manual "README.md"))
```
Tag or hash is not mandatory. If hash is missing it asks if you want to continue. If tag is missing then main branch is cloned and used. Do you think it should ask about that too? "Git branch is not set, cloning main. Continue? y/n"

If both git and url exists then git is preferred.

I tried to use as much already existing "pipeline" as possible, so this code clones the git repository to temporary directory. Then it packages the repository into .tgz and installs it.

I'm not sure if spec '() is okay, on line 2564.

This also needs more testing but I'm making this pull request because I need it for context here https://github.com/ashinn/snow-fort/pull/14.


Retropikzel:
Thank you!

PR #1044, opened 2025-09-12
Snow chibi command to install program dependencies

Retropikzel:
I was thinking of something like pip freeze that would make a file with program dependencies in it. But then I realised why have another file when program has their dependencies in it. So I added a small command to install programs dependencies.

main.scm:

```
(import (scheme base)
        (scheme write)
        (retropikzel hello))

(hello)
```

Command:
```
snow-chibi install-dependencies main.scm
```

Installs (retropikzel hello).

I do think the name of the command is too long. Do you have suggestions for a better one?

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -637,10 +637,10 @@
       (close-output-port out))))
 
 (define (command/install-dependencies cfg spec scm-file)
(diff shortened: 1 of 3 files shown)


ashinn:
I think `install-dependencies` is fine, can't think of a better alternative.

PR #1059, opened 2025-12-11
When building Kawa libraries add the install dir to loadpath

Retropikzel:
Small fix for something I overlooked when adding Kawa support.


Retropikzel:
Updated pull requst from latest chibi master branch.

Retropikzel:
Thank you!

PR #1071, opened 2026-01-08
Git index fixes

Retropikzel:
Github repos did not index correctly before, this fixes it.


Retropikzel:
The problem still exists, previous commit just fixed the name to have question mark at the end.

PR #1073, opened 2026-01-11
Git index improvements

Retropikzel:
Show error when

- In non git directory
- Non package .tgz file
- Empty git repository, which means hash is "HEAD"


Retropikzel:
Thank you!

PR #1079, opened 2026-01-31
Improve Kawa library path finding

Retropikzel:
Kawa binary sets first item of CLASSPATH to be, on default install, /usr/local/share/lib/kawa/kawa.jar. So the library install path can be taken from there. It now works even when Kawa is not installed with default configuration.

Additionally --install-prefix option is also added before the install path.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -1550,22 +1550,23 @@
             path
             "/usr/local/share/guile/"))))
     ((kawa)
-     (make-path
-       (if (conf-get cfg 'install-prefix) (conf-get cfg 'install-prefix) "")
-       (path-directory
-         (car
-           (string-split
-             (call-with-temp-file
-               "snow-kawa.scm"
-               (lambda (tmp-path out preserve)
-                 (with-output-to-file
-                   tmp-path
-                   (lambda ()
-                     (display "(import (scheme base) (scheme write) (scheme process-context))")
-                     (newline)
-                     (display "(display (get-environment-variable \"CLASSPATH\"))")))
-                 (process->string `(kawa ,tmp-path))))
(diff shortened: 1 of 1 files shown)


lib/chibi/snow/commands.scm
```diff
@@ -1550,7 +1550,23 @@
...
+               (call-with-temp-file
+                 "snow-kawa.scm"
+                 (lambda (tmp-path out preserve)
+                   (with-output-to-file
+                     tmp-path
+                     (lambda ()
+                       (display "(import (scheme base) (scheme write) (scheme process-context))")
+                       (newline)
+                       (display "(display (get-environment-variable \"CLASSPATH\"))")))
+                   (process->string `(kawa ,tmp-path))))
```
ashinn: Just

```
(process->string `(kawa -e "(display (get-environment-variable \"CLASSPATH\"))"))
```

should work.  Also try to fall back to something reasonable if this command fails.

lib/chibi/snow/commands.scm
```diff
@@ -1553,20 +1553,17 @@
...
-                       (display "(import (scheme base) (scheme write) (scheme process-context))")
-                       (newline)
-                       (display "(display (get-environment-variable \"CLASSPATH\"))")))
-                   (process->string `(kawa ,tmp-path))))
-               #\:))))))
+         (let ((kawa-classpath
+                 (string-split
+                   (process->string
+                     `(kawa -e "(display (get-environment-variable \"CLASSPATH\"))"))
+                   #\.)))
```
ashinn: Splitting on dot won't work, e.g. I have /kawa-3.1.1/ in my path.  How about splitting on : and dropping the kawa.jar?

Retropikzel:
Thank you!

Issue #1060, opened 2025-12-12
SRFI-64 test-error stops program when there is an error instead of passing the test


Retropikzel:
```
(import (scheme base)
        (scheme write)
        (srfi 64))


(test-begin "SRFi-64")

(test-error (begin (error "Test should pass")))

(display "And the program should continue")
(newline)

(test-end "SRFi-64")
```

Should report 1 passed test, insted it errors and stops.

Retropikzel:
I'm not familiar with how the SRFI-64 reference implementation works inside. Thats why I opened issue instead of pull request. But here is pull request now: https://github.com/ashinn/chibi-scheme/pull/1061

One problem is that snow-chibi will not install another SRFI-64 implementation over Chibi's native one, which is of course intended and good. But the test-error not working and eq? being eq crash my tests. The pull request might not be a perfect fix but now the test programs should atleast finish. :)

Retropikzel:
I forgot to mention this issue on the pull request. https://github.com/ashinn/chibi-scheme/pull/1080. It fixes this so I'm closing the issue.

PR #1080, opened 2026-02-08
When cond-expanding for srfi-9 or srfi-34 also or for r7rs

Retropikzel:
Updated Chibi's SRFI-64 library to have same changes that were merged into the original https://github.com/scheme-requests-for-implementation/srfi-64/pull/11


Retropikzel:
Thank you!

Retropikzel:
Mentioning the issue here for documentation purposes. https://github.com/ashinn/chibi-scheme/issues/1060

PR #1081, opened 2026-02-09
Tr7 program command fixes

Retropikzel:
Sometimes the file can be #f, this takes it into consideration.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -1744,7 +1744,7 @@
              `(stklos -A ,install-dir ,file)))
         ((tr7)
          (if lib-path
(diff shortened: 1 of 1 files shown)


lib/chibi/snow/commands.scm
```diff
@@ -1757,7 +1757,7 @@
              `(stklos -A ,install-dir ,file)))
         ((tr7)
          (if lib-path
-             `(sh -c ,(string-append "TR7_LIB_PATH=" lib-path " tr7i " file))
+             `(sh -c ,(string-append "TR7_LIB_PATH=" lib-path " tr7i " (if file file "")))
```
ashinn: Not introduced by this change, but doesn't the file need quoting?

Retropikzel:
Actually I need to double check this

PR #1082, opened 2026-02-10
Guile path improvements

Retropikzel:
This make Guile code work the same way as others. The compiled files are placed next to .sld and .scm files.

Before this change this command:
```
snow-chibi --impls=guile install --install-library-dir=snow --install-source-dir=snow retropikzel.hello
```
Tries to install to system directories. Same with:
```
snow-chibi --impls=guile install --install-prefix=snow retropikzel.hello
```


Retropikzel:
Thank you!

Issue #1078, opened 2026-01-31
Snow-chibi installs SRFI even when Chibi has it


Retropikzel:
```
cd /tmp
snow-chibi install --prefix=snow --impls=chibi srfi.39
tree snow/
snow/
└── share
    └── snow
        └── chibi
            └── srfi
                ├── 39.scm
                └── 39.sld
```

I dont think snow-chibi should install SRFI if implementation already supports it? In this case Chibi also uses SRFI-39 in a way that installing the one from snow-fort will break chibi resulting into this kind of error:

```
WARNING: exception inside module with no imports - did you forget to (import (scheme base)) in: (scheme base)
ERROR: module attempted to reference itself while loading: (scheme base)
  called from <anonymous> on line 210 of file /usr/local/share/chibi/meta-7.scm
  called from <anonymous> on line 1188 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 804 of file /usr/local/share/chibi/init-7.scm
  called from error on line 108 of file /usr/local/share/chibi/init-7.scm
  called from for1 on line 81 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 804 of file /usr/local/share/chibi/init-7.scm
  called from call-with-current-continuation on line 827 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 259 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 250 of file /usr/local/share/chibi/meta-7.scm
  called from <anonymous> on line 279 of file /usr/local/share/chibi/init-7.scm
  called from for1 on line 81 of file /usr/local/share/chibi/init-7.scm
  called from for1 on line 81 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 259 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 250 of file /usr/local/share/chibi/meta-7.scm
  called from <anonymous> on line 279 of file /usr/local/share/chibi/init-7.scm
  called from for1 on line 81 of file /usr/local/share/chibi/init-7.scm
  called from for1 on line 81 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 259 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 250 of file /usr/local/share/chibi/meta-7.scm
  called from <anonymous> on line 437 of file /usr/local/share/chibi/meta-7.scm
[...]

PR #1083, opened 2026-02-12
Make SRFI test program writing more reliable

Retropikzel:
This pull request is related to #1078. It does not fix the whole thing but is a step into that direction.

Part of the problem for example with Kawa is that sometimes the SRFI checking code is not fully written into the tempfile.

This is weird as it is Chibi who is doing the writing.

Old implementation here: https://github.com/ashinn/chibi-scheme/blob/da20079bf5b1fbc40be622e9c3e23456665f1476/lib/chibi/snow/commands.scm#L1894

First putting the strings into list and then writing them out one by one seems more reliable than writing things out in a loop inside `with-output-to-file`.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -1894,26 +1894,31 @@
 (define (native-srfi-support impl cfg)
   (letrec*
     ((max-srfis 500)
+     (srfi-conds '())
      (gen-srfis-conds
        (lambda (count)
-         (when (< count max-srfis)
-           (display "  ")
-           (display
-             (string-append
-               "(cond-expand ((or "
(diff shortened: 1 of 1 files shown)


Retropikzel:
Current main:
`snow-chibi install --impls=kawa srfi.1`
> Can't find package: (srfi 1)

Tempfile writing fails, thinks Kawa does not support any SRFI's.

With changes in this pull request.

`snow-chibi install --impls=kawa srfi.1` silently not installed.

`snow-chibi install --impls=kawa srfi.39` installed, as there is candidate available. Will be address in another pull request.

Retropikzel:
Thank you!

PR #1084, opened 2026-02-12
Do not install natively supported libraries

Retropikzel:
Related to #1078 

- Ignore native libraries even when snow-fort package exists
- Add warning when not installing SRFI that is natively supported

Correct me if I am wrong but now that the SRFI support check bug is fixed I think the `(null? candidates)` is a bug. Because the behaviour before these changes was:
`snow-chibi install --impls=kawa srfi.1` is not installed because there is no package in snow-fort.
`snow-chibi install --impls=kawa srfi.39` is  installed because there is package in snow-fort.

And after these changes neither is installed, and warning is issued on both occasions that they are skipped.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -3014,10 +3014,11 @@
             (if (member (car ls) lib-names)
                 (warn "skipping already installed library: " (car ls)))
             (lp (cdr ls) res (cons (car ls) ignored)))
-           ((and (null? candidates)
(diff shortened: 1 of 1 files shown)


Retropikzel:
@ashinn You said in https://github.com/ashinn/chibi-scheme/issues/1085 that you will try to do release soon. I have Skint support in https://github.com/Retropikzel/chibi-scheme/tree/skint which is more urgent to get to any release than this. So I'm closing this pull request and bringing the skint support in first.

PR #1086, opened 2026-02-19
Add support for Skint

Retropikzel:


lib/chibi/snow/commands.scm
```diff
@@ -1748,6 +1756,10 @@
          (if lib-path
              `(racket -I r7rs -S ,install-dir -S ,lib-path --script ,file)
              `(racket -I r7rs -S ,install-dir --script ,file)))
+        ((skint)
+         (if lib-path
+             `(stklos -A ,install-dir -A ,lib-path ,file)
```
ashinn: stklos?

Retropikzel: Sorry, typo caused by working with so many implementations. I will fix it.

Retropikzel:
Thank you!

PR #1087, opened 2026-02-21
Fix string-append problem with tr7 when file is #f

Retropikzel:
Since tr7 has string-append in scheme-program-command, when file is #f there is an error. This fixes that.


Retropikzel:
Thank you!

PR #1092, opened 2026-03-20
Kawa fixes

Retropikzel:
When running Kawa inside Docker without -it flags which do this
```
  -i, --interactive                      Keep STDIN open even if not attached
  -t, --tty                              Allocate a pseudo-TTY
```

Installing for example SRFI-180 from here https://github.com/srfi-explorations/r7rs-srfi for running tests does not work. srfi/180.sld does not exists in the tmp directory. I'm not sure why it happens and what is causing it. Sometimes with Kawa I noticed I needed to use (newline) to get output to appear. So maybe it is something related to that.

Anyway, this fixes it for some reason. Not a perfect fix but I dont have time to do a thorough investigation now.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -2642,12 +2642,13 @@
 
 (define (kawa-builder impl cfg library dir)
   (let* ((src-library-file (make-path dir (get-library-file cfg library)))
-         (res (system 'kawa
-                      (string-append "-Dkawa.import.path="
-                                     (car (get-install-dirs impl cfg)))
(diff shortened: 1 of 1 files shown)


Retropikzel:
To document this more it seems to happend when library has dependencies. In the mentioned SRFI repo it does not happen with SRFI-60 or SRFI-145. SRFI-180 is packaged with snow-chibi and then indexed locally and installed.

Retropikzel:
Okay this happens with other implementations too. I'm closing this pull request and filing an issue when I can make a repeatable test for this.

PR #1095, opened 2026-03-22
Meevax support for snow-chibi

Retropikzel:


lib/chibi/snow/commands.scm
```diff
@@ -1563,6 +1563,11 @@
                           (- (string-length (car kawa-classpath)) 8)))))))
     ((loko)
      (list "/usr/local/share/r6rs"))
+    ((meevax)
+     (list
+       (make-path
+         (car (string-split (process->string '(meevax "--library-directories"))
```
ashinn: You can use `process->string-list` (lines).

Retropikzel: Changed it to use process->string-list.

Retropikzel:
Thank you!

Issue #1097, opened 2026-03-26
Snow-chibi to allow git repository as main repository


Retropikzel:
Currently only siblings are allowed to be git repositories. But something like this should be made to work:
`snow-chibi --repo=https://codeberg.org/retropikzel/snow-chibi-testing.git search foo`

The .snow-repo.scm file in the git repository should be used as main repository.

Retropikzel:
https://github.com/Retropikzel/chibi-scheme/tree/snow-chibi-repo-arg-git-support Adds support for git urls to be used with --repo

Retropikzel:
Turns out the support for --host being git repository the changes from the branch mentioned is previous comment are needed. So this https://github.com/Retropikzel/chibi-scheme/tree/snow-chibi-allow-git-host branch contains that too.

In addition it errors if user tries to upload package when host is git repository. And of course allows using git repository as main host.

Retropikzel:
https://github.com/ashinn/chibi-scheme/pull/1099

Retropikzel:
The pull request has been merged so I'm closing this.

PR #1099, opened 2026-03-29
Snow chibi allow git host

Retropikzel:
Related issue https://github.com/ashinn/chibi-scheme/issues/1097

- Allow --repo urls be git urls
- Allow --host urls be git urls
- Throw an error and stop if user tries to register key or upload a package when host is git url

Https git urls of these repositories where used for testing:
https://codeberg.org/Snow-fort/snow-fort
https://gitlab.com/snow-fort/snow-fort

```
snow-chibi --host=https://codeberg.org/snow-fort/snow-fort.git search foo
> (foo bar)	1.0.3
> (foo bar)	1.0.2
> (foo bar)	1.1.0
```

```
snow-chibi --host=https://gitlab.com/snow-fort/snow-fort.git search hello
> (retropikzel hello)	1.2.3
```

but of course any user repository works as host, but then naturally only packages from that repository are used.

```
snow-chibi --host=https://codeberg.org/retropikzel/snow-chibi-testing.git search foo
> (foo bar)	1.0.3
> (foo bar)	1.0.2
snow-chibi --host=https://codeberg.org/retropikzel/snow-chibi-testing.git search hello
> No libraries matched your query.
```


lib/chibi/snow/commands.scm
```diff
@@ -933,6 +933,8 @@
         (newline)))))
 
 (define (command/reg-key cfg spec)
+  (when (string-suffix? ".git" (remote-uri cfg '(command package uri) ""))
```
ashinn: Can we factor out an `is-git-repo-uri?` or similar utility?

ashinn (commented):
Thanks!

lib/chibi/snow/commands.scm
Retropikzel: Thats a good idea. I'll add it.

Retropikzel: Added this.

Retropikzel:
No also gen-key command will error if host url is git. As it does not make much sense either to use in that situation.

Retropikzel:
Thank you!

Issue #1098, opened 2026-03-26
Snow-chibi show only one entry per package version in search results


Retropikzel:
Currently if the repository has multiple versions of the same package, the package will be listed multiple times in search results. It should only be shown once.

Retropikzel:
Note to myself that there is a fix in https://github.com/Retropikzel/chibi-scheme/tree/search-duplicate-fix waiting for https://github.com/ashinn/chibi-scheme/pull/1095 to be merged first.

It's just a simple fix for now. It will not work if multiple authors release package with same name. But that is propably something that has to be dealt with in the future anyway. Involving more thought, planning and discussions.

Retropikzel:
I immediately realised that there is already packages with same name. For example (srfi 64). Code is now updated to show authors.

Retropikzel:
Seems like it will not work in situation like this, this is before the fixes when all versions are shown. The fix assumes that packages are listed biggest version number first. But it seems they are not.
```
snow-chibi --host=https://gitlab.com/snow-fort/snow-fort.git search foo
(foo bar)	1.0.3
(foo bar)	1.0.2
(foo bar)	1.1.0
```

Libraries need to be sorted first by version. Or version needs to be check on the loop.

Retropikzel:
Fixed.

PR #1100, opened 2026-04-02
Do not show duplicates on search results. Format search result output

Retropikzel:
Fix for issue https://github.com/ashinn/chibi-scheme/issues/1098

Edit:
- Add package-id to package library


lib/chibi/snow/commands.sld
```diff
@@ -31,6 +31,7 @@
           (srfi 1)
           (srfi 27)
           (srfi 95)
+          (srfi 166 columnar)
```
ashinn, line 35: Maybe not always safe to use (srfi 166) formatters with (chibi show), see if you can replace the (chibi show) import with (srfi 166).

lib/chibi/snow/commands.scm
```diff
@@ -1174,10 +1176,36 @@
 ;; Prints a list of libraries whose meta-info contain any of the given
 ;; keywords.  Returns in sorted order for how well the package matches.
 
-(define (summarize-libraries cfg lib-names+pkgs)
-  (for-each (lambda (name pkg) (describe-library cfg name pkg))
-            (map car lib-names+pkgs)
-            (map cdr lib-names+pkgs)))
+(define (summarize-libraries cfg lib-names+pkgs . repo)
+  (let ((repository (if (null? repo) '(repository) (car repo)))
+        (sorted-lib-names+pkgs (sort lib-names+pkgs string>=? package-version))
```
ashinn:
```suggestion
        (sorted-lib-names+pkgs
         (delete-duplicates
          (sort lib-names+pkgs string>=? package-version)
          (lambda (a b) (equal? (package-id (cdr a)) (package-id (cdr b))))))
```

where

```
(define (package-id pkg) (cons (package-name pkg) (package-author pkg)))
```

lib/chibi/snow/commands.scm
```diff
@@ -1174,10 +1176,36 @@
...
+                (let ((pkg-identifier (cons (package-name pkg)
+                                            (package-author repository pkg))))
+                  (when (not (member pkg-identifier added-identifiers))
+                    (set! added-identifiers pkg-identifier)
+                    (set! results (cons pkg results)))))
+              (map car sorted-lib-names+pkgs)
+              (map cdr sorted-lib-names+pkgs))
+    (show
+      #t
+      (columnar
```
ashinn: Hmmm... fine for now but makes me think there should be an array formatter utility so you can just collect into an array for cases like this.

ashinn (commented):
Thanks!

lib/chibi/snow/commands.scm
Retropikzel: I'll change it to this.

I think at some point lib/chibi/package.sld/scm needs somekind of identification functions. There is also package-hash in snow-fort pull request https://github.com/Retropikzel/snow-fort/blob/master/pkg-page/index.scm#L26. Perhaps `(define (package-id pkg . exclude-version?) ...)`. Because I feel that `(equal? (package-id pkg-a) (package-id pkg-b))` when pkg-a and pkg-b are same package but different version should return false. But in practice it might not be a problem.

lib/chibi/snow/commands.sld
Retropikzel: Switched to (srfi 166) from (chibi show).

lib/chibi/snow/commands.scm
Retropikzel: I added package-id to lib/chibi/snow/package.sld/scm. @ashinn can you check if it's allright? By default it includes the version of the package, but it can be excluded. I figured that when it is used the user should know if they want version into it. But if it is used for example checking if package is installed, then we want version into it. I'll add mention of it into the pull request description too.

lib/chibi/snow/commands.scm
```diff
@@ -1174,10 +1176,40 @@
...
+(define (summarize-libraries cfg lib-names+pkgs . repo)
+  (let ((repository (if (null? repo) '(repository) (car repo)))
+        (sorted-lib-names+pkgs
+         (delete-duplicates
+          (sort lib-names+pkgs string>=? package-version)
+          (lambda (a b) (equal? (package-id repo (cdr a) #t)
+                                (package-id repo (cdr b) #t)))))
+        (added-identifiers '())
+        (results '()))
+    (for-each (lambda (name pkg)
```
ashinn: With the delete-duplicates you no longer need this block at all.

Retropikzel: Removed. Also cleaned up the code.  To test package with multiple versions you can do `snow-chibi --use-curl --host=https://gitlab.com/snow-fort/snow-fort.git search foo`, `(foo bar)` has versions 1.0.2, 1.0.3 and 1.1.0 in the git repositories. The snow-fort main repository only contains one version of each library.

lib/chibi/snow/package.scm
```diff
@@ -101,6 +101,15 @@
            (any library-name (package-libraries package))
            (any program-name (package-programs package)))))
 
+(define (package-id repo pkg . exclude-version?)
+  (append (list (package-name pkg)
+                (package-author repo pkg)
```
ashinn: So the original intention was that packages would be unique by publisher + library name (choosing the broadest name if there are multiple libs in the pkg).  The signatures, trust mechanisms, and download paths are based off of publishers.  It's conceivable for multiple people to repackage the same code by a separate author.  Maintainer should probably always be synonymous with publisher but I wouldn't rely on that.

So I think we can replace both author and maintainer here with package-publisher (which needs to be defined).

Retropikzel: Packages in git repositories do not have publishers, only the one that are uploaded in the "traditional" way. I think the publisher field is added during upload?

Mandatory --publisher flag could be added to package command?

ashinn: I think we can infer the publisher from the git uri (e.g. for `https://github.com/user/foo` use `user@github.com`).  We shouldn't trust someone to specify the publisher themselves.

Retropikzel: Thats a good idea. Should I add and switch to (package-publisher ...) in this pull request and make new issue for the git-index command adding publisher to the .snow-repo.scm?

Edit: Just as I posted this I realised the publisher does not need to be added as a field if (git (url ...)) exists, it can be read from there. I think it all can fit into this pull requests scope.

Retropikzel: Added package-publisher and also fixed git-index command.

lib/chibi/snow/package.scm
```diff
@@ -110,6 +118,23 @@
...
+          ((assq 'git pkg)
+           (let* ((uri (string->uri (package-git-url pkg #t)))
+                  (host (uri-host uri))
+                  (repository-name (path-directory (uri-path uri))))
+             (string-append (string-copy repository-name 1) "@" host)))
+          ((repo-find-publisher repo email)
+           => (lambda (pub)
+                (let ((name (assoc-get pub 'name)))
+                  (if name
+                    (string-append name " <" (or email "") ">")
```
ashinn: I think we can just use the email and ignore the name here.

lib/chibi/snow/package.scm
```diff
@@ -110,6 +118,23 @@
 (define (strip-email str)
   (string-trim (regexp-replace '(: "<" (* (~ (">"))) ">") str "")))
 
+(define (package-publisher repo pkg)
+  (let ((email (package-email pkg)))
+    (or (cond
+          ((assq 'git pkg)
```
ashinn: nitpick: Just in case it's a badly formatted git package this is safer:

```suggestion
          ((package-git-url pkg #t) => (lambda (uri) ...)
```

ashinn (commented):
Thanks!  Almost there :)

lib/chibi/snow/package.scm
Retropikzel: So removing lines 130->135? This is copied from package-author so I must admin I dont understand those lines fully. The idea was that package-publisher = get it from git url or fall back to similar behaviour than package-author.

lib/chibi/snow/package.scm
Retropikzel: Thanks!  I changed it to that.

lib/chibi/snow/package.scm
ashinn: `package-author` is mostly for display purposes.  `package-publisher` we're using for identification so it should be the raw email string.  Something like:

```
(define (package-publisher repo pkg)
  (cond
    ((package-git-url pkg #t)
     => (lambda (uri)
          (let* ((uri (string->uri (package-git-url pkg #t)))
                 (host (uri-host uri))
                 (repository-name (path-directory (uri-path uri))))
            (string-append (string-copy repository-name 1) "@" host))))
    (else (package-email pkg))))
```

Retropikzel: Thank you!

Retropikzel:
@ashinn Updated to the latest master branch and should be good to go?

Retropikzel:
Thank you!

PR #1107, opened 2026-05-23
Add SRFI 253 support

aartaka:
Hello! This adds support for SRFI 253. Tell me if something is out of order!

Is there maybe some Chibi API for type checking? If so, I’d like to use that instead of sample implementation. But I haven’t found any mention of types in the docs, so…


ashinn:
> Hello! This adds support for SRFI 253. Tell me if something is out of order!
> 
> Is there maybe some Chibi API for type checking? If so, I’d like to use that instead of sample implementation. But I haven’t found any mention of types in the docs, so…

There's `define-binary-record-type` whose purpose is serialization but includes type enforcement as a side effect.  Otherwise the convention in Chibi code is to use `(chibi assert)`.

I don't automatically include all SRFI's in Chibi these days, but this is small enough and is supported by both Chicken and Gauche so it's OK.

aartaka:
@ashinn thanks for your work and for the explanation!

PR #1109, opened 2026-05-26
More reliable Chicken version checking

Retropikzel:
The version checking is broken on Chicken 6. So I changed it to something more reliable.


Retropikzel:
Thank you!

Issue #1111, opened 2026-05-26
Improving snow-chibi handling missing config file


Retropikzel:
Currently if `~/.snow/config.scm` does not exist an error message is shown. Like so:

```
retropikzel:~$ snow-chibi update
couldn't load config:: "/home/retropikzel/.snow/config.scm"
ERROR in "open-input-file": couldn't open input file: "/home/retropikzel/.snow/config.scm"
  called from conf-load on line 171 of file /usr/local/share/chibi/chibi/config.scm
  called from <anonymous> on line 235 of file /usr/local/bin/snow-chibi.scm
  called from <anonymous> on line 1268 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 800 of file /usr/local/share/chibi/init-7.scm
```

I assume this is quite confusing to new users. And I have to admit it annoys me too to have create this file in every dockerfile. I know it can be ignored but ignoring ERROR messages does not feel right.

Would it be okay to either
A) Remove this message completely
B) On first run write a config file with all the fields but commented out and display message like "Default config file written to ~/.snow/config.scm"
This could improve usability by still telling new users that there is a config file and where it is. And by having the options in there already with explanations. Like so:

```
(
    ;(option1 value1) ;; This changes X
    ;(option2 value2) ;; This changes y
)
```
C) Display only `couldn't load config:: "/home/retropikzel/.snow/config.scm`
D) Write empty list into the config file silenty
E) Something else?

ashinn:
C sgtm

PR #1112, opened 2026-05-27
Do not show full error message when config file does not exist

Retropikzel:
Continuation of https://github.com/ashinn/chibi-scheme/issues/1111

What do you think about this? I added `(file does not exist)` so that the message is still there. But now it is a warning instead of error and only one line. It might be a bit weird way to do it but I did not want to change the (chibi config) library.

### DIFF
--- a/tools/snow-chibi.scm
+++ b/tools/snow-chibi.scm
@@ -5,6 +5,7 @@
 
 (import (scheme base)
         (scheme process-context)
+        (scheme file)
         (chibi snow commands)
         (chibi snow interface)
         (chibi app)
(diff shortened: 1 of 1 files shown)


tools/snow-chibi.scm
```diff
@@ -231,6 +232,13 @@
...
-                 (command-line)
-                 (conf-load (conf-default-path "snow")))
+(run-application
+  app-spec
+  (command-line)
+  (let ((conf-file (conf-default-path "snow")))
+    (cond ((file-exists? conf-file)
+           (conf-load conf-file))
+          (else
+            (warn "couldn't load config (file does not exist)" conf-file)
```
ashinn:
```suggestion
            (warn "snow config file not found: " conf-file)
```

tools/snow-chibi.scm
```diff
@@ -231,6 +232,13 @@
...
+            (parameterize ((current-error-port (open-output-string)))
+              (conf-load conf-file))))))
```
ashinn: This will always fail and return the default.  I should make the params optional, but you can just use:

```
(make-conf '() #f #f (current-second))
```

Retropikzel:
Thank you!

PR #1120, opened 2026-06-07
Allow square brackets as paren aliases under -B CLI option

aartaka:
Hi! 

This is a first shot at parsing square brackets as aliases for parens as discussed in #883. I’ve opted for pretty simplistic strategy of offloading all parsing to () cases instead of introducing any custom logic. Which makes something like `(+ 1 2]` valid syntax, but preserves simplicity and codebase size. Is that a fine tradeoff? 

Contrived (and horrible) test case:

``` scheme
(let ([x 1]
      [hello (string-append "hello" " you")])
  (string-append hello " " [number->string x]))
```


aartaka:
CC @ashinn

sexp.c
```diff
@@ -3428,6 +3423,9 @@ sexp sexp_read_raw (sexp ctx, sexp in, sexp *shares) {
   case '"':
     res = sexp_read_string(ctx, in, '"');
     break;
+  case '[':
+    if (sexp_global(ctx, SEXP_G_ALLOW_SQUARE_BRACKETS) != SEXP_TRUE)
```
ashinn: Elegant, but not what's discussed in the ticket.  The point is to make this configurable.  Also you need to update the SRFI 38 implementation in the same way (and ideally the scribble reader as well).

aartaka: Yes, I realize that this diverges from the ticket’s suggestion, but I prioritized having the feature earlier rather than later. Addition of `$bracket-list$` (that isn’t that configurable in Kawa either) can happen later if there’s a need for it.

I’ll update SRFI 38. What exactly should change in scribble reader?

aartaka: @ashinn re SRFI 38: is that fine if square brackets are hardcoded into the parser? I’m not sure how to approach configurability there. 

Another question: does it actually have to belong to SRFi 38? 38 specifies external representation portable across implementations, and even allowing square brackets there makes it non-portable. Is that right?

ashinn: I'm of the personal stance that it's a huge waste of precious reader characters to dedicate square brackets to the same purpose as parenthesis, so would not accept a non-configurable patch.  There should be a port-specific configuration, coming from a global default.  If it's a symbol, that symbol is prefixed to the list, if `#t`, there is no prefix, if `#f` then square brackets raise an exception.

Regarding SRFI 38, it should follow the same configuration.  This will likely require a new utility in (chibi ast), which should be cond-expanded on import, just using `#f` for non-Chibi impls.  Square brackets are not valid syntax in SRFI 38 so it's a valid extension, though not very friendly for people who want to write portable code (nor are square brackets in general).

aartaka: @ashinn so okay, let me summarize the way I understood it. Here’s your quote from the original issue #883:

> Yes, runtime is preferable. We can consider:
>
> 1. global setting (configured via command-line option and/or code)
> 2. per-extension setting (e.g. by default .sls would follow R6RS and .scm would follow Kawa)
> 3. per-file setting (via #! pragmas and/or allowing the library declaration language to customize for include files)
>
> 1 is simplest for now so I recommend that unless you feel ambitious.

From this, I inferred that CLI flag is an acceptable option. As someone with limited resources, I decided to go for it, as probably the only one I could possibly implement. It’s also consistent with the way CHICKEN works, which to me already feels like a proven good practice.

> Yes, all of these are ultimately port-level settings (like case-insensitivity), it's just a question of where the defaults come from.

While this is true, I consider this option quite complex and expect this to take a lot of effort. Again, I’m aiming for the Chibi sentiment of staying small and neat, thus this PR size.

Now to your comment here:

> I'm of the personal stance that it's a huge waste of precious reader characters to dedicate square brackets to the same purpose as parenthesis, so would not accept a non-configurable patch. 

It is configurable already, via a CLI flag. While not particularly pretty and composable, it’s enough as per your comment in #883. So I opted for it.

Another point can be that Scheme reader in general is quite primitive and non-extensible, so occupying a reader char is much safer than e.g. in Common Lisp, where _any_ char can be a user-defined reader syntax and thus collide with e.g. square brackets.

> There should be a port-specific configuration, coming from a global default. If it's a symbol, that symbol is prefixed to the list, if #t, there is no prefix, if #f then square brackets raise an exception.

~This design is expressed in this PR, except for the part with port-specificity, for reasons stated above.~ Sorry, misread. This is not the design expressed in PR, but other points still stand.

> Regarding SRFI 38, it should follow the same configuration. This will likely require a new utility in (chibi ast), which should be cond-expanded on import, just using #f for non-Chibi impls. Square brackets are not valid syntax in SRFI 38 so it's a valid extension, though not very friendly for people who want to write portable code (nor are square brackets in general).

Portability is a huge selling point for me, so adding a non-portable non-standard (albeit widely agreed-on) syntax to an inherently standard-abiding (as per SRFI nature) code sound… wrong?

ashinn: > [...]
> From this, I inferred that CLI flag is an acceptable option. As someone with limited resources, I decided to go for it, as probably the only one I could possibly implement. It’s also consistent with the way CHICKEN works, which to me already feels like a proven good practice.

Yes, just a CLI flag is totally fine, and preferred to start (sorry I forgot to clarify that).  That flag should be configurable though - it should allow specifying a symbol.  That's the only thing I'm objecting to.

> Another point can be that Scheme reader in general is quite primitive and non-extensible, so occupying a reader char is much safer than e.g. in Common Lisp, where _any_ char can be a user-defined reader syntax and thus collide with e.g. square brackets.

My argument is not about safety, but utility.  I've stated this many times in other forums, I won't repeat the arguments here.

> > Regarding SRFI 38, it should follow the same configuration. This will likely require a new utility in (chibi ast), which should be cond-expanded on import, just using #f for non-Chibi impls. Square brackets are not valid syntax in SRFI 38 so it's a valid extension, though not very friendly for people who want to write portable code (nor are square brackets in general).
> 
> Portability is a huge selling point for me, so adding a non-portable non-standard (albeit widely agreed-on) syntax to an inherently standard-abiding (as per SRFI nature) code sound… wrong?

`(scheme read)` uses the SRFI 38 implementation, not the C implementation.  If you object to extending `(scheme read)` then you need to refactor a common, configurable core Scheme reader, and have SRFI 38 always explicitly disable square brackets.

aartaka: > > From this, I inferred that CLI flag is an acceptable option. As someone with limited resources, I decided to go for it, as probably the only one I could possibly implement. It’s also consistent with the way CHICKEN works, which to me already feels like a proven good practice.
> 
> Yes, just a CLI flag is totally fine, and preferred to start (sorry I forgot to clarify that). That flag should be configurable though - it should allow specifying a symbol. That's the only thing I'm objecting to.

What do you mean by specifying a symbol here? Like choosing a paren alias symbol?

> > Another point can be that Scheme reader in general is quite primitive and non-extensible, so occupying a reader char is much safer than e.g. in Common Lisp, where _any_ char can be a user-defined reader syntax and thus collide with e.g. square brackets.
> 
> My argument is not about safety, but utility. I've stated this many times in other forums, I won't repeat the arguments here.

I see. Well, I have no other responses to that but personal preference. I’ve seen it used a lot for increased readability and prettiness, and, working with Clojure, I also grew to like square brackets. So yeah, I personally like that, like many other Lisp programmers. Is that enough?

> > > Regarding SRFI 38, it should follow the same configuration. This will likely require a new utility in (chibi ast), which should be cond-expanded on import, just using #f for non-Chibi impls. Square brackets are not valid syntax in SRFI 38 so it's a valid extension, though not very friendly for people who want to write portable code (nor are square brackets in general).
> > 
> > 
> > Portability is a huge selling point for me, so adding a non-portable non-standard (albeit widely agreed-on) syntax to an inherently standard-abiding (as per SRFI nature) code sound… wrong?
> 
> `(scheme read)` uses the SRFI 38 implementation, not the C implementation. If you object to extending `(scheme read)` then you need to refactor a common, configurable core Scheme reader, and have SRFI 38 always explicitly disable square brackets.

Oh, I see now! I’ll amend SRFI 38 then!

aartaka: Oh wait, now I get it. By symbol you mean something like `-B '$bracket-list$'` forcing `[a b c] -> ($bracket-list$ a b c)`, right?

ashinn: Yes, exactly.  `-B#t` would then parse `[a b c]` as `(a b c)`.  `-B#f` would be the current default behavior, with square brackets as identifier characters (not exceptions as I mistakenly suggested above).

You might want a shorthand for `-B#t` which is ugly, maybe `-B.`.  Another option is `-6` which could later be extended to set other R6RS defaults.

aartaka: What about having `-B` for `(a b c)` and `-B sym` / `-Bsym` for `(sym a b c)`? Or is that too ambiguous for CLI argument parsing?

aartaka: And then, `-B#f` is not needed, as omitting `-B` does that already?

ashinn: We do this for `-R`, which runs a specified module's `main`, defaulting to `(chibi repl)`.  We can do this because it's mutually incompatible with specifying a file argument to `chibi-scheme`.  We can't do this with `-B` because there's no way to distinguish whether `chibi-scheme -B foo` means parse `[a b c]` as `(foo a b c)`, or parse it as `(a b c)` and load the file `foo`.

aartaka: Yes, adding space is a bad decision, but having an unspaced thing (`-Bfoo` and `-B`) should be unambiguous, right?

ashinn: But it's inconsistent.  All other options which take arguments have an optional space.

aartaka: I see. `#t` and `#f` it is then!

aartaka:
@ashinn I’ve pushed a new version of sexp.c parser, but I’m kinda lost in regards to SRFI 38 reader. Can you please suggest the detailed-ish course of action here? I have to admit that I didn’t fully understand the sentence below:

> This will likely require a new utility in (chibi ast), which should be cond-expanded on import, just using #f for non-Chibi impls.

ashinn:
In `read-one` in `read-with-shared-structure`, you need to check for `#\[`.  The behavior will depend on `SEXP_G_SQUARE_BRACKETS_SYM`.  Since that's in C, you'll need to make it accessible to Scheme as an opcode.  Something like:

```
sexp sexp_square_brackets_sym (sexp ctx, sexp self, sexp_sint_t n) {
  return sexp_global(ctx, SEXP_G_SQUARE_BRACKETS_SYM);
}
```

I guess this is small enough to put in opcodes.c instead of ast.c:

```
_FN0(_I(SEXP_OBJECT), "square-brackets-symbol", 0, sexp_square_brackets_sym),
```

aartaka:
@ashinn I believe it’s implemented now.

ashinn:
Thank you!

aartaka:
> Thank you!

Thanks for guiding me :black_heart: 

I seem to have forgotten a couple of newlines in the CLI help menu, can you add them?

ashinn:
> I seem to have forgotten a couple of newlines in the CLI help menu, can you add them?

Done.

PR #1121, opened 2026-06-08
Support for net on windows

kmatveev:
Support for network on Windows via winsock.

Implementation details:
On Unix sockets are just files, but Windows is a different story, sockets are special type of handles. Sockets-specific calls require such handle. However, Windows provides _open_osfhandle() which wraps socket into file descriptor which supports read() and write() calls. In order to support both socket-specific operations and general I/O operations we need to store both socket and file descriptor. So I've added new field "sock" to fileno. On Windows fd and sock are different, on Unix they are the same.

I've tested my changes via:
1. examples/echo-server-udp.scm
2. a similar single-threaded echo server for tcp which uses ports with read-line and display

### DIFF
--- a/include/chibi/sexp.h
+++ b/include/chibi/sexp.h
@@ -512,7 +512,9 @@ struct sexp_struct {
     struct {
       char openp, no_closep;
       sexp_sint_t fd, count;
+#ifdef _WIN32      
       SOCKET_TYPE sock;
+#endif      
     } fileno;
     struct {
       sexp kind, message, irritants, procedure, source, stack_trace;
@@ -1272,7 +1274,11 @@ enum sexp_uniform_vector_type {
 #define sexp_fileno_fd(f)        (sexp_pred_field(f, fileno, sexp_filenop, fd))
 #define sexp_fileno_count(f)     (sexp_pred_field(f, fileno, sexp_filenop, count))
 #define sexp_fileno_openp(f)     (sexp_pred_field(f, fileno, sexp_filenop, openp))
+#ifdef _WIN32
 #define sexp_fileno_sock(f)      (sexp_pred_field(f, fileno, sexp_filenop, sock))
+#else
+#define sexp_fileno_sock(f)      (sexp_port_fd(f))
+#endif
 #define sexp_fileno_no_closep(f) (sexp_pred_field(f, fileno, sexp_filenop, no_closep))
 
 #define sexp_ratio_numerator(q)   (sexp_pred_field(q, ratio, sexp_ratiop, numerator))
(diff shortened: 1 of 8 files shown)


include/chibi/sexp.h
```diff
@@ -18,6 +18,14 @@ extern "C" {
 #include "chibi/features.h"
 #include "chibi/install.h"
 
+#if !(defined _WIN32) || defined(__CYGWIN__)
+#include <sys/socket.h>
+#define SOCKET_TYPE sexp_sint_t
+#else
+#include <winsock2.h>
```
kmatveev, line 25: Windows wants winsock.h to be included before windows.h, so I've added it here, and also moved non-windows socket.h here to make a single if/else

lib/chibi/net.stub
```diff
@@ -34,41 +40,44 @@
...
 
 ;;> Returns a list of 2 new sockets, the input and output end of a new
 ;;> pipe, respectively.
-
-(define-c errno (open-socket-pair "socketpair")
-  (int int int (result (array fileno 2))))
+(cond-expand
+  (windows)
+  (else 
+    (define-c errno (open-socket-pair "socketpair")
```
kmatveev, line 79: there are no socketpair on windows, and on Unix it is mostly used with "UNIX" and "LOCAL" sockets, so it is not portable anyway

ashinn:
Thank you for working on this @kmatveev!  I'll review as soon as I get a chance.

include/chibi/sexp.h
```diff
@@ -507,6 +512,7 @@ struct sexp_struct {
     struct {
       char openp, no_closep;
       sexp_sint_t fd, count;
+      SOCKET_TYPE sock;
```
ashinn, line 516: This is redundant on Unix, how about:

```suggestion
#ifdef _WIN32
      SOCKET_TYPE sock;
#endif
```

and then below:

```
#ifdef _WIN32
#define sexp_fileno_sock(f)      (sexp_pred_field(f, fileno, sexp_filenop, sock))
#else
#define sexp_fileno_sock(f)      (sexp_port_fd(f))
#endif
```

sexp.c
```diff
@@ -206,7 +206,23 @@ sexp sexp_finalize_port (sexp ctx, sexp self, sexp_sint_t n, sexp port) {
   if (sexp_port_openp(port)) {
     sexp_port_openp(port) = 0;
     if (sexp_oportp(port)) sexp_flush_forced(ctx, port);
-#ifndef PLAN9
+#ifdef _WIN32
```
ashinn: This code is identical to the Unix version, except for the constants `SD_BOTH`, `SD_RECEIVE` and `SD_SEND` which we can factor out definitions for, and for `sexp_port_sock` which using the above suggestion we can also use on Unix.

sexp.c
```diff
@@ -1974,6 +1990,31 @@ sexp sexp_make_fileno_op (sexp ctx, sexp self, sexp_sint_t n, sexp fd, sexp no_c
   return res;
 }
 
+sexp sexp_make_fileno_sock (sexp ctx, sexp self, sexp_sint_t fd, SOCKET_TYPE sock) {
```
ashinn: This is almost identical to the above `sexp_make_fileno` except for the one line `sexp_fileno_sock(res) = sock;`.  You can just call `sexp_make_fileno` here for the body, and then add that one line.  Also, this is only used by accept.c so it can be moved there.

ashinn (changes requested):
Thanks!

include/chibi/sexp.h
kmatveev: I agree, will do

kmatveev:
I've updated PR with suggested changes

ashinn:
Thanks!

Issue #1093, opened 2026-03-21
Installing package from file does not install dependencies


Retropikzel:
main.scm
```
(import (scheme base)
        (scheme write)
        (foo bar))
(hello)
```

foo/bar.sld
```
(define-library
  (foo bar)
  (import (scheme base)
          (retropikzel hello))
  (export hello))
``` 

```
snow-chibi package foo/bar.sld
snow-chibi install --impls=chibi foo-bar.tgz
chibi-scheme main.scm
> ERROR: couldn't find import: (retropikzel hello)
  called from resolve-import on line 156 of file /usr/local/share/chibi/meta-7.scm
  called from <anonymous> on line 168 of file /usr/local/share/chibi/meta-7.scm
  called from for1 on line 81 of file /usr/local/share/chibi/init-7.scm
  called from for1 on line 81 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 259 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 250 of file /usr/local/share/chibi/meta-7.scm
  called from <anonymous> on line 437 of file /usr/local/share/chibi/meta-7.scm
  called from <anonymous> on line 1268 of file /usr/local/share/chibi/init-7.scm
  called from <anonymous> on line 800 of file /usr/local/share/chibi/init-7.scm
```

Retropikzel:
Fix for this has been merged. Closing.

PR #1123, opened 2026-06-09
Install dependencies when installing from file

Retropikzel:
When installing package from file also install it's dependencies. Fix for https://github.com/ashinn/chibi-scheme/issues/1093

I had to move the immediate fail when package does not exist from `expand-package-dependencies` into `command/install`. So that when given nonexisting package as argument the fail is immediate, but when dependency does not exist user is prompted to continue.

```
snow-chibi install --impls=chibi lol.lel
WARNING: snow config file not found: : "/home/retropikzel/.snow/config.scm"
Can't find package: (lol lel)
```

From the issue.

main.scm:
```
(import (scheme base)
        (scheme write)
        (foo bar))
(hello)
```

foo/bar.sld:
```
(define-library
  (foo bar)
  (import (scheme base)
          (retropikzel hello))
  (export hello))
```

```
snow-chibi package foo/bar.sld
snow-chibi install --impls=chibi foo-bar.tgz
chibi-scheme main.scm
Hello
```

Add nonexisting dependency to foo/bar.sld:
```
(define-library
  (foo bar)
  (import (scheme base)
          (does not exist)
          (retropikzel hello))
  (export hello))
```

```
snow-chibi package foo/bar.sld
snow-chibi install --impls=chibi foo-bar.tgz
WARNING: skipping already installed library: : (scheme base)
Can't find package: (does not exist).  Proceed anyway? [y/n]: 
```

Currently it will output extra warnings for `(scheme ...)` libraries. This is because another bug. Related to `implementation-supports-natively?` and checks. Which I plan to fix next.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -3080,9 +3080,8 @@
             ;; assume certain core libraries already installed
             ;; (info "assuming core library installed: " (car ls))
             (lp (cdr ls) res (cons (car ls) ignored)))
-           ((and (null? candidates) (member (car ls) lib-names))
-            (die 2 "Can't find package: " (car ls)))
-           ((null? candidates)
+           ((or (null? candidates)
+                (and (null? candidates) (member (car ls) lib-names)))
             (cond
              ((yes-or-no? cfg "Can't find package: " (car ls)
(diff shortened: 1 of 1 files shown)


Retropikzel:
Thank you!

Issue #1124, opened 2026-06-12
Fails to build with gcc 14


Retropikzel:
Steps to replicate with docker. Builds started failing today. Worked okay couple hours ago.
Latest commit on git log:
```
commit 78f93ce102b8e9998ff91ecf2a4e350253800695 (HEAD -> master, origin/master, origin/HEAD)
Merge: 6fd23611 7e65499c
Author: Alex Shinn <alexshinn@gmail.com>
Date:   Fri Jun 12 21:07:45 2026 +0900

    Merge pull request #1121 from kmatveev/windows-net
    
    Support for net on windows
```

Dockerfile.chibi:
```
FROM debian:trixie-slim
RUN apt-get update && apt-get install -y gcc git ca-certificates make
COPY chibi-scheme /chibi-scheme
WORKDIR /chibi-scheme
RUN gcc --version
RUN make
```

Run:
```
git clone https://github.com/ashinn/chibi-scheme.git
docker build -f Dockerfile.chibi .
```

Does not compile. Here is the the output.
```
#9 0.208 echo '#define sexp_so_extension "'.so'"' > include/chibi/install.h
#9 0.223 echo '#define sexp_default_module_path "'/usr/local/share/chibi:/usr/local/lib/chibi:/usr/local/share/snow:/usr/local/lib/snow'"' >> include/chibi/install.h
#9 0.226 echo '#define sexp_platform "'linux'"' >> include/chibi/install.h
#9 0.230 echo '#define sexp_architecture "'x86_64'"' >> include/chibi/install.h
#9 0.233 echo '#define sexp_version "'0.12.0'"' >> include/chibi/install.h
#9 0.236 echo '#define sexp_release_name "'`cat RELEASE`'"' >> include/chibi/install.h
#9 0.242 cc -c  -Iinclude  -DSEXP_USE_NTPGETTIME -DSEXP_USE_INTTYPES -Wall -g -g3 -O3  -o main.o main.c
#9 0.912 cc -c  -Iinclude  -DSEXP_USE_NTPGETTIME -DSEXP_USE_INTTYPES -Wall -g -g3 -O3  -fPIC -o gc.o gc.c
#9 1.179 cc -c  -Iinclude  -DSEXP_USE_NTPGETTIME -DSEXP_USE_INTTYPES -Wall -g -g3 -O3  -fPIC -o sexp.o sexp.c
#9 1.255 In file included from sexp.c:5:
[...]

PR #1125, opened 2026-06-12
Fixed retrieval of socket from fileno

kmatveev:
It was tricky to test: these defines are used as left side in assignments and as right side in retrievals. So in UDP echo server a socket fd value was written instead of fileno pointer, and this value was later retrieved during send() and receive(), which worked as long as no other fields were written/read.

### DIFF
--- a/Makefile.detect
+++ b/Makefile.detect
@@ -192,7 +192,7 @@ endif
 
 CHIBI_POSIX_COMPILED_LIBS = lib/chibi/process$(SO) lib/chibi/time$(SO) \
 	lib/chibi/system$(SO) lib/chibi/stty$(SO) lib/chibi/pty$(SO) \
-	lib/chibi/net$(SO) lib/srfi/18/threads$(SO)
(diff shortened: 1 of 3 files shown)


kmatveev:
This should fix issue #1124

Makefile.detect
```diff
@@ -192,7 +192,7 @@ endif
 
 CHIBI_POSIX_COMPILED_LIBS = lib/chibi/process$(SO) lib/chibi/time$(SO) \
 	lib/chibi/system$(SO) lib/chibi/stty$(SO) lib/chibi/pty$(SO) \
-	lib/chibi/net$(SO) lib/srfi/18/threads$(SO)
```
kmatveev, line 195: In my previous PR I've removed net from CHIBI_POSIX_COMPILED_LIBS in Makefile, and missed CHIBI_POSIX_COMPILED_LIBS in Makefile.detect

ashinn:
Thanks for the fix!

Issue #1127, opened 2026-06-16
Support SRFI 235 (combinators)?


aartaka:
Would you be open to (my) contribution adding SRFI 235 to Chibi? I personally think that this is the most useful SRFI as it makes composing many things quite effortless. And thus it deserves place in the core. Any objections?

Retropikzel:
@aartaka If youre interested there is https://github.com/srfi-explorations/r7rs-srfi and https://github.com/srfi-explorations/srfi-test in which I have been adding SRFI's into snow-fort. If you add SRFI-235 there it can be tested on multiple implementations and trough snow-fort kind of added to all of them.

It's a bit of work in progress in terms of someone else contributing into it but I can improve it based on your feedback. And any other portable SRFI is of course welcome there too.

aartaka:
@Retropikzel this sounds nice, although I’m indecisive if I should pursue this. It’d be better to have it right in Chibi, after all. @ashinn what do you think?

ashinn:
@aartaka I'm interested in promoting snow-fort (or any other package managers), specifically so things like this don't have to be packaged in the core.

aartaka:
@ashinn so what’s the course of action then?

ashinn:
Let me close this for now.

aartaka:
Thanks for resolving it!

Retropikzel:
@aartaka Just popping in to let you know I packaged SRFI-235 and it's now in snow-fort.

aartaka:
@aartaka Just popping in to let you know I packaged SRFI-235 and it's now in snow-fort.


Oh, nice, thank you!

PR #1128, opened 2026-06-16
Add SRFI 235

aartaka:
Sorry @ashinn, I went ahead with this despite no input in #1127. Feel free to close if you don’t feel like adding this.


aartaka:
Note that I’m not particularly attached to this PR now that I reviewed the SRFI again. Most of the combinators seem useless to me, so I’m okay with just copy-pasting the necessary ones in my code, instead of bloating Chibi core.

PR #1130, opened 2026-06-23
Fix logic error in indexing duplicate package removal

Retropikzel:


Retropikzel:
That was fast! Thanks. :)

PR #1133, opened 2026-06-26
fixed problem with bignum reminders

mbamac:
This is fix for bignum reminders sings. This is not my solution. It was generated by Claude AI, so if you want to use it, please review it before.


ashinn:
Is there a test case this fixes?

mbamac:
this patch fixes failures from bignum division tests (from make test-all).

ashinn:
Thanks!  Claude is right about the root cause (negating bignums in place that we don't own), but I pushed a fix with cleaner code in 43136bb1e417f3b4813e19ab004668e1be2f0785.

mbamac:
superb 👍🏻

PR #1134, opened 2026-07-04
Racket install improvements

Retropikzel:
The old version would (import (scheme base)) and reimporting something from there would then lead to an error. This version writes the complete content of .sld file into .rkt file.


mxork:
I think in-lining the entire file isn't necessary here. `include` is available in the initial `#lang r7rs` module environment, so the intitial `(import (scheme base))` can just be removed.

mxork:
Though, I am curious about the original error, since the spec only says:

> In a program or library declaration, it is an error to import the same identifier more than once with different bindings, or to redefine or mutate an imported binding with a definition or with set!, or to refer to an identifier before it is imported.

So I'm assuming you hit the "same identifier with different bindings case"? Or something weirder coming out of #lang r7rs

Retropikzel:
> I think in-lining the entire file isn't necessary here. `include` is available in the initial `#lang r7rs` module environment, so the intitial `(import (scheme base))` can just be removed.

Thanks! I didnt know that. I changed the code to that.

> Though, I am curious about the original error, since the spec only says:
> 
> > In a program or library declaration, it is an error to import the same identifier more than once with different bindings, or to redefine or mutate an imported binding with a definition or with set!, or to refer to an identifier before it is imported.
> 
> So I'm assuming you hit the "same identifier with different bindings case"? Or something weirder coming out of #lang r7rs

Something like that. I can not remember where I got the error anymore. I should have included it in the pull request.

Retropikzel:
Thank you!

PR #1138, opened 2026-07-10
Add emscripten build github action

Retropikzel:
This is related to an issue so I opened it even when I have another pull request open.


Retropikzel:
Added both cmake and make js emscripten builds. Assuming that pull request https://github.com/ashinn/chibi-scheme/pull/1144 will be merged at some point. If not I can make changes.

Retropikzel:
Latest emsdk is installed in the build and now it works. These commits can be squashed if you want to make things bit cleaner.

Retropikzel:
Thank you!

PR #1140, opened 2026-07-10
srfi 18: time->seconds does not round to nearest second, returns exact

mxork:
The implementation of time->seconds rounds the representation of the `time?` to the nearest second. The srfi seems pretty clear that this isn't the intended behavior:

```
(time->seconds time)                                  ;procedure
Converts the time object time into an exact or inexact real number representing the number of seconds elapsed since some implementation dependent reference point.

    (time->seconds (current-time))  ==>  955039784.928075   ;; <------
```

It also makes the srfi's interface unusable for timeouts <1s, which is goofy.

My change does the obvious thing. It uses an exact (rational) representation for microseconds, because I think that `(seconds->time (time->seconds T))` should equal `T`. That said, using an inexact? to match `current-second` from `(scheme time)` seems tasteful as well.

### DIFF
--- a/lib/srfi/18/interface.scm
+++ b/lib/srfi/18/interface.scm
@@ -3,10 +3,15 @@
 (define (time? x) (timeval? (if (pair? x) (car x) x)))
 
 (define (time->seconds x)
-  (timeval-seconds (if (pair? x) (car x) x)))
+  (let ((t (if (pair? x) (car x) x)))
+    (+ (timeval-seconds t)
+       (* 1e-6 (timeval-microseconds t)))))
 
 (define (seconds->time x)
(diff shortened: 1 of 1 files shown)


ashinn:
Thanks!  Unless there is precedent for using exact rationals, I think it's better to use inexact here for better compatibility.  Chibi can also be compiled without rational support.

mxork:
Inexact does seem like the right thing, doesn't it. Patched, and this output looks sensible:

```
(import (scheme small) (srfi 18))

(define now (car (current-time)))
(define rt (seconds->time (time->seconds now)))

(display now)
(newline)
(display rt)

; output:
; #<timeval 140171138348800>
; #<timeval 140171138352096>
```

PR #1141, opened 2026-07-14
use monotonic, higher resolution jiffies

mxork:
The current-jiffy was 1:1 with current-second, but assuming that its meant to be used more for measuring intervals and not wall time, I replaced it with more suitable clocks:

1. posix uses [CLOCK_MONOTONIC](https://www.man7.org/linux/man-pages/man3/clock_gettime.3.html). an argument could be made for CLOCK_MONOTONIC_RAW on linux.
2. windows uses [QueryPerformanceCounter](https://learn.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancecounter)
3. plan9 keep using `nsec`. they have [`cycles`](https://github.com/brho/plan9/blob/master/sys/include/libc.h#L345), but its not a clock out of the box. 

I haven't tested the windows or plan9 code, but it looks sane to me. Advice on cross-compiling from linux to windows appreciated. I couldn't bring myself to install mingw under wine, but I also couldn't get `make CC=mingw... PLATFORM=windows` to Just Work.

### DIFF
--- a/lib/scheme/time.c
+++ b/lib/scheme/time.c
@@ -119,9 +119,9 @@ sexp sexp_current_jiffy (sexp ctx, sexp self, sexp_sint_t n) {
 #if SEXP_MS_JIFFY
   LARGE_INTEGER frequency;
   QueryPerformanceFrequency(&frequency);
-  return sexp_make_fixnum((1000*t)/frequency);
+  return sexp_make_fixnum((1000*t.QuadPart)/frequency.QuadPart);
 #else
-  return sexp_make_fixnum(t);
+  return sexp_make_fixnum(t.QuadPart);
 #endif
 
 #elif !defined(PLAN9)
@@ -157,7 +157,7 @@ sexp sexp_jiffies_per_second(sexp ctx, sexp self, sexp_sint_t n) {
 #ifdef _WIN32
   LARGE_INTEGER frequency;
   QueryPerformanceFrequency(&frequency);
-  return sexp_make_fixnum(frequency);
+  return sexp_make_fixnum(frequency.QuadPart);
 #elif !defined(PLAN9)
   return sexp_make_fixnum(1000000000);
 #else
--- a/lib/scheme/time.c
+++ b/lib/scheme/time.c
@@ -4,12 +4,15 @@
 /* BSD-style license: http://synthcode.com/license.txt       */
 
 #include <chibi/eval.h>
+#include <chibi/sexp.h>
+#include <stdint.h>
 
 #ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
 #elif !defined(PLAN9)
 #include <sys/time.h>
+#include <time.h>
 #else
 typedef long time_t;
 #endif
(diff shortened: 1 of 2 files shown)


lib/scheme/time.c
```diff
@@ -108,11 +111,44 @@ sexp sexp_current_clock_second (sexp ctx, sexp self, sexp_sint_t n) {
...
+  LARGE_INTEGER t;
+  QueryPerformanceCounter(&t);
+  return sexp_make_fixnum(t);
+#elif !defined(PLAN9)
+  struct timespec tv;
+  int err = clock_gettime(CLOCK_MONOTONIC, &tv);
+  // :note could use errno, here
+  if (err)
+    return sexp_user_exception(ctx, self, "couldn't get current jiffy", SEXP_FALSE);
+  return sexp_make_fixnum(1000000000*tv.tv_sec + tv.tv_nsec);
```
ashinn: Hmmm... on a 32-bit system, this overflows after the system has been up for 4 seconds.

You could check and make a bignum as needed, but that becomes always on 32-bit systems.  Probably better to conditionally define `jiffies-per-second` as a smaller value, say millisecond granularity, on 32-bit systems.

mxork: Yep. I'll polish up a patch for it tomorrow.

A millisecond is an awfully long time. But nothing better to be done within the constraints of the spec.

ashinn: Keep using chibi, you'll come to think a millisecond is fast! :wink:

mxork: patched and pushed. hopefully I didn't whiff an ifdef.

mxork:
I don't know why the CMAKE builds decided to all explode; not clear its from my change?

lib/scheme/time.c
```diff
@@ -108,11 +111,68 @@ sexp sexp_current_clock_second (sexp ctx, sexp self, sexp_sint_t n) {
...
+
+sexp sexp_jiffies_per_second(sexp ctx, sexp self, sexp_sint_t n) {
+#if SEXP_MS_JIFFY
+    return sexp_make_fixnum(1000);
+#else
+#ifdef _WIN32
+  LARGE_INTEGER frequency;
+  QueryPerformanceFrequency(&frequency);
+  return sexp_make_fixnum(frequency.QuadPart);
+#elif !defined(PLAN9)
```
ashinn: This is the same as the next branch, you can unify them.

lib/scheme/time.c
```diff
@@ -108,11 +111,68 @@ sexp sexp_current_clock_second (sexp ctx, sexp self, sexp_sint_t n) {
...
+#endif
+
+#elif !defined(PLAN9)
+  struct timespec tv;
+  int err = clock_gettime(CLOCK_MONOTONIC, &tv);
+  // :note could use errno, here
+  if (err)
+    return sexp_user_exception(ctx, self, "couldn't get current jiffy", SEXP_FALSE);
+  uint64_t current_ns = 1000000000*tv.tv_sec + tv.tv_nsec;
+#if SEXP_MS_JIFFY
```
ashinn: This block is repeated below, you can move it out of the `current_ns` computation just once at the end.

ashinn (commented):
Thanks!

lib/scheme/time.c
mxork: ack.

lib/scheme/time.c
mxork: ack

ashinn:
Thanks!

PR #1143, opened 2026-07-16
Add parse-stream-file-index to (chibi parse)

cstawarz:
I've written a parser with `(chibi parse)`, and I need its output to include detailed location information (source file name, line, and column) for various entities, to be used for error reporting when the parsed program executes. To do this, I need to be able to convert a local index in a specific parse stream to a global index in the source file. These changes enable this by adding a `file-offset` field to `Parse-Stream` and exporting a new procedure, `parse-stream-file-index`.


ashinn:
Thanks! LGTM.

cstawarz:
Thanks!

PR #1149, opened 2026-07-28
Snow chibi list all

Retropikzel:
Adds a list-all command to snow-chibi. It has options:

- --sexp?
  - Output as Scheme readable s-expression
- --dotted-format?
  - Library names in dotted format, (retropikzel tap) -> retropikzel.tap
- --separator
  - Set the separator for library list items

Examples:
```
snow-chibi list-all
> (retropikzel tap) (srfi 170) (retropikzel compile-r7rs)...
snow-chibi -sexp?=1 list-all
> ((retropikzel tap) (srfi 170) (retropikzel compile-r7rs)...)
snow-chibi --dotted-format list-all
> retropikzel.tap srfi.170 retropikzel.compile-r7rs...
snow-chibi --sexp?=1 --dotted-format?=1 list-all
> (retropikzel.tap srfi.170 retropikzel.compile-r7rs...)
snow-chibi --separator="___" list-all
> (retropikzel tap)___(srfi 170)___(retropikzel compile-r7rs)...

### DIFF
--- a/tools/snow-chibi.scm
+++ b/tools/snow-chibi.scm
@@ -95,7 +95,8 @@
     (foreign-depends (list string) "foreign libraries library depends on")
     (use-curl? boolean ("use-curl") "use curl for file uploads")
     (sexp? boolean ("sexp") "output information in sexp format")
-    ))
+    (dotted-format? boolean "output the package names in dotted format")
+    (separator string "string to separate package names with, defaults to space")))
 
 (define (conf-default-path name)
   (or (get-environment-variable "SNOW_CHIBI_CONFIG")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: tools/snow-chibi.scm [hunk 1]
```diff
@@ -180,6 +187,9 @@
      (search
       "search for packages"
       (@ ,@search-spec) (,command/search terms ...))
+     (list-all
```
ashinn, line 190: Currently `search` with no terms just takes a long time to output nothing.  We could repurpose that to be equivalent to `list-all`, though it may not be intuitive.  It's fine to make `list-all` a separate command, but they should share the same underlying code.

ashinn (commented):
Thanks!

tools/snow-chibi.scm
Retropikzel: The idea did not occur to me that empty search could do the same. The list-all command is something I need for my projects so I could use empty search instead. I'm not very fond of "polluting" the main arguments list with things that I have added there so I would rather remove list-all command and make empty search output all libraries. I will close this pull request and open a new one to not pollute the commit history.

Retropikzel:
Adding link here for documentation purposes. https://github.com/ashinn/chibi-scheme/pull/1155

PR #1152, opened 2026-08-02
Use generics in SRFI 253 define-checked

aartaka:
Re-using generic infra (that I didn’t know of earlier) seems to be better than reinventing it. Quite a minimal change too!

I found several bugs in the process, and had to comment out parts of SRFI 253 tests due to that:
- Invoking the checked record constructor right after `define-record-type-checked` seems to error out. Running it after some time has passed does not error.
- Generics don’t support rest args. Is there a reason for that? I’ll file a pull request fixing that in an hour.

### DIFF
--- a/lib/srfi/253.sld
+++ b/lib/srfi/253.sld
@@ -3,7 +3,8 @@
           (scheme case-lambda)
           (scheme list)
           (chibi assert)
-          (chibi ast))
+          (chibi ast)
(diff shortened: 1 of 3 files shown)


aartaka:
CC @ashinn

aartaka:
@ashinn and what about complementary #1153?

PR #1153, opened 2026-08-03
Allow variadic generics

aartaka:
This allows defining variadic generics, bridging a previously puzzling gap between regular procedures and generics. This incurs a search cost. Technically quadratic, practically negligible (I doubt anyone defines as many methods as necessary to make this stutter.)

### DIFF
--- a/lib/chibi/generic.scm
+++ b/lib/chibi/generic.scm
@@ -32,15 +32,26 @@
   (er-macro-transformer
    (lambda (e r c)
      (let ((name (car (cadr e)))
-           (params (map (lambda (param)
-                          (if (identifier? param)
-                              `(,param (lambda _ #t))
-                              param))
-                        (cdr (cadr e))))
+           (params (let lp ((params (cdr (cadr e))))
+                     (cond
+                      ((not (pair? params)) '())
+                      ((identifier? (car params))
+                       (cons `(,(car params) (lambda _ #t))
+                             (lp (cdr params))))
+                      (else
+                       (cons (car params) (lp (cdr params)))))))
+           (variadic? (let lp ((params (cdr (cadr e))))
+                        (cond
+                         ((pair? params) (lp (cdr params)))
+                         ((null? params) #f)
+                         (else params))))
            (body (cddr e)))
        `(,(r 'generic-add!) ,name
          (,(r 'list) ,@(map cadr params))
-         (,(r 'lambda) (,(r 'next) ,@(map car params))
+         ',variadic?
+         (,(r 'lambda) ,@(if variadic?
+                             `((,(r 'next) ,@(map car params) . ,variadic?))
+                             `((,(r 'next) ,@(map car params))))
           (,(r 'let-syntax) ((call-next-method
                               (,(r 'syntax-rules) ()
                                ((_) (,(r 'next))))))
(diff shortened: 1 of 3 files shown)


lib/chibi/generic.scm
```diff
@@ -63,28 +77,41 @@
...
+                                   (lambda (x) boolean?)
                                    procedure?)
-                             (lambda (next t p f)
-                               (set! methods (insert-method! methods p f))))))
+                             #f
+                             (lambda (next tag preds rest? func)
+                               (set! methods (insert-method! methods preds rest? func))))))
     (lambda args
-      (let ((len (length args)))
+      ;; Yes, this is quadratic, walking all arities in order. This is
```
ashinn: Sorry, but this is not acceptable.  If you can find a way to make only the variadic cases slower I might consider that.

aartaka:
Alright, I pushed a new version that has linear search for required args
and only searches variadics when these are exhausted. @ashinn please
take another look.

lib/chibi/generic.scm
```diff
@@ -59,26 +70,58 @@
 
 ;;> Create a new first-class generic function named \var{name}.
 
+(define (sublist lst capped-length)
```
ashinn: take?

lib/chibi/generic.scm
```diff
@@ -32,15 +32,26 @@
...
+                      ((identifier? (car params))
+                       (cons `(,(car params) (lambda _ #t))
+                             (lp (cdr params))))
+                      (else
+                       (cons (car params) (lp (cdr params)))))))
+           (variadic? (let lp ((params (cdr (cadr e))))
+                        (cond
+                         ((pair? params) (lp (cdr params)))
+                         ((null? params) #f)
+                         (else params))))
```
ashinn: We don't use this as a symbol, do we?  Let's just return `#t` here.

lib/chibi/generic.scm
```diff
@@ -59,26 +70,58 @@
...
   (let ((name name)
-        (methods (make-vector 6 '())))
+        (methods (make-vector 6 '()))
+        (variadic-methods (make-vector 6 '())))
     (vector-set! methods
-                 3
+                 4
                  (list (cons (list (lambda (x) (eq? x add-method-tag))
                                    (lambda (x) (list? x))
+                                   (lambda (x) (or (boolean? x)
```
ashinn, line 91: After changing the variadic? to `#t` in the true case we can just use `boolean?` (no lambda) here.  Also above, that could actually just be written as `list?` without a lambda.

lib/chibi/generic.scm
```diff
@@ -100,5 +143,5 @@
 ;;> that applies when all parameters match the given list
 ;;> of predicates \var{preds}.
```
ashinn: Let's update the documentation:

```
;;> Extend the generic \var{g} with a new method \var{f}
;;> that applies when all parameters match the given list
;;> of predicates \var{preds}, or if `variadic?` is true a prefix
;;> of the parameters matches \var{preds}.
```

lib/chibi/generic.scm
```diff
@@ -59,26 +70,58 @@
...
+                               (if rest?
+                                   (set! variadic-methods (insert-method! variadic-methods preds func))
+                                   (set! methods (insert-method! methods preds func)))))))
     (lambda args
-      (let ((len (length args)))
+      (letrec* ((len (length args))
+                (search-variadic
+                 (lambda (idx)
+                   (cond
+                    ((> idx len)
```
ashinn:
```suggestion
                    ((>= idx (vector-length variadic-methods))
```

ashinn (changes requested):
Thanks!

lib/chibi/generic.scm
aartaka: Do you mean renaming this one or using the SRFI 1 procedure? I’ll assume the former, as using SRFIs in core Chibi is probably not an option

lib/chibi/generic.scm
aartaka: It’s used below, when the lambda list is constructed. see `if variadic?` true branch.

lib/chibi/generic.scm
ashinn: I meant rename.  It's not forbidden to use SRFIs here, but for something this small I think it's fine to copy.

lib/chibi/generic.scm
ashinn: OK, but then it's probably better to call this `rest-param`.

lib/chibi/generic.scm
aartaka: No, this `>` is deliberate. Try this:

``` scheme
(define-generic hello)

(define-method (hello . args)
  args)

(hello)
```

ashinn: Without this check, the `(vector-ref variadic-methods idx)` below can throw out of bounds.

aartaka: And the condition is right:
<dl>
  <dt> (> idx len)
  <dd> If the number of required args is more than provided
  <dt> (>= idx (vector-length variadic-methods))
  <dd> If there are no more variadic methods
</dl>

We need the former, because the only condition for variadic matching failing is when there are more required arguments than what was provided. I.e.

``` scheme
(define-generic hello)

(define-method (hello a b c . rest)
  a)

(hello 1)
```

We cannot simply compare the arity (= length of method vector,) because methods are variadic and thus extend over required arity length.

aartaka:
@ashinn I think I addressed your comments and responded to those needing extra deliberation.

lib/chibi/generic.scm
aartaka: > Without this check, the (vector-ref variadic-methods idx) below can throw out of bounds.

Oh, right, indeed! Fix pushed.

lib/chibi/generic.scm
aartaka: Renamed!

lib/chibi/generic.scm
aartaka: Okay, I renamed it.

lib/chibi/generic.scm
aartaka: @ashinn there was indeed an out-of-bounds, as tested on
``` scheme
(define-generic hello)

(define-method (hello a b c)
  (list a b c))

(hello 1 2 3 4 5 7 8 9 2 3 32 2)
```

aartaka:
@ashinn thanks!

PR #1155, opened 2026-08-05
If no search keywords are given, list all libraries

Retropikzel:
If no search keywords are given, skip calculating the score and give everyone score of 100.


Retropikzel:
Thank you!

PR #1156, opened 2026-08-05
Message when installing library and it has no tests

Retropikzel:
The branch is very poorly named. But basically when checking the test output with automation this helps. When the tests all have passed the output will contain text "All tests passed.", when the library has not tests it will contain text "Library has no tests.".

Often when library has no tests the output is not empty, usually there is atleast one  `WARNING: skipping already installed library: : <some library>`

If you do not like it I can move it behind the `verbose?` flag.


Retropikzel:
Thank you!

PR #1157, opened 2026-08-06
More output on verbose mode. Gambit test running fixes

Retropikzel:
Two changes but hopefully small enough that it's okay. Gambit was not running tests correctly so I added more output on verbose mode. And then it was quite easy to copy the command and test it with different file.


Retropikzel:
Gambit installs packages to `~/.gambit_userlib` so to test you can run:
```
rm -rf ~/.gambit_userlib/retropikzel/hello.*
snow-chibi install --impls=gambit --always-yes --install-tests?=1 --show-tests?=1 --verbose?=1 "(retropikzel hello)"
```

Retropikzel:
Thank you!

PR #1160, opened 2026-08-08
Include the non-matching data in the "no matching pattern" error

aartaka:
The “no matching pattern” error not mentioning the mismatched data brings a lot of pain to me, due to using `match` a lot on diverse inputs. I have to insert extra error clauses just to avoid this error.

``` scheme
[something-else
 (error "No match: " something else)]
```

Listing the offending data in “no matching pattern” clause seems relatively intuitive, thus the patch.

Am I missing something? Why wasn’t it present before?

### DIFF
--- a/lib/chibi/match/match.scm
+++ b/lib/chibi/match/match.scm
@@ -333,7 +333,7 @@
   (syntax-rules (=>)
     ;; no more clauses, the match failed
     ((match-next v g+s)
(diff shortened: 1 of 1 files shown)


ashinn:
> Listing the offending data in “no matching pattern” clause seems relatively intuitive, thus the patch.
> 
> Am I missing something? Why wasn’t it present before?

No reason, thanks for the patch!

PR #1161, opened 2026-08-08
Stricter SRFI 273 define-values-checked

aartaka:
Sorry, I introduced a bug in the previous PR—`define-values-checked` was defining vars _before_ checking them. Now it aborts the definition on mismatch, as it should.


ashinn:
> Sorry, I introduced a bug in the previous PR—`define-values-checked` was defining vars _before_ checking them. Now it aborts the definition on mismatch, as it should.

Could you add a unit test for this?

aartaka:
@ashinn added.

PR #1162, opened 2026-08-10
Fix test running on Guile, Mosh and Racket

Retropikzel:
After these fixes all implementations except mit-scheme run tests on (retropikzel hello). Meaning that the test program starts and displays hello. So the test running itself works.


Retropikzel:
Thank you!

PR #1164, opened 2026-08-11
Fix test-package command to work on all implementations

Retropikzel:
I forgot my test value into the previous code, so any test run was run with Chibi. 

- Fixes that
- Allows to use the multiple --impls=chibi,chicken,gauche form
- verbose?  output has implementation too
- Add always-yes? true when running tests, otherwise if the tests fail snow-chibi will ask do you want to continue

Quick way to test:
```
cd /tmp
wget https://snow-fort.org/s/iki.fi/retropikzel/srfi/64/0.2.1/srfi-64-0.2.1.tgz
snow-chibi test-package --impls=chibi --verbose?=1 srfi-64-0.2.1.tgz
snow-chibi test-package --impls=cyclone --verbose?=1 srfi-64-0.2.1.tgz
```

The tests will fail on Cyclone, which is expected.

### DIFF
--- a/lib/chibi/snow/commands.scm
+++ b/lib/chibi/snow/commands.scm
@@ -654,16 +654,20 @@
       (close-output-port out))))
 
 (define (command/test-package cfg spec pkg-file)
-  (call-with-temp-dir
-    "pkg-test"
-    (lambda (dir preserve)
-      (let* ((pkg (package-file-meta pkg-file))
-             (snowball (maybe-gunzip (file->bytevector pkg-file)))
-             (test-cfg
-               (conf-extend cfg
-                            (list '(command (test-package (show-tests? . #t)))))))
-        (tar-extract snowball (lambda (f) (make-path dir (path-strip-top f))))
-        (test-package 'chibi test-cfg pkg dir)))))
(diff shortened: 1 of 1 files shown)


lib/chibi/snow/commands.scm
```diff
@@ -654,16 +654,21 @@
...
+          (lambda (dir preserve)
+            (let* ((pkg (package-file-meta pkg-file))
+                   (snowball (maybe-gunzip (file->bytevector pkg-file)))
+                   (test-cfg
+                     (conf-extend cfg
+                                  (list '(command (test-package (show-tests? . #t)))
+                                        '(always-yes? . #t)))))
+              (tar-extract snowball (lambda (f) (make-path dir (path-strip-top f))))
+              (test-package impl test-cfg pkg dir)))))
+      (if (list? impls) impls (list impls)))))
```
ashinn:
```suggestion
      (conf-get-list cfg 'implementations '(chibi)))))
```

and remove the let impls above.

Retropikzel: Yes, that is better. I'll change it to that. I did not notice that conf-get-list exists.

Retropikzel:
Thank you!

PR #1167, opened 2026-08-13
Preserve minimum possible arity in case-lambda-generated procedure

aartaka:
It always grated me that `case-lambda`-created procedures had `0+` arity. Even when having actual clear arglists with fixed / minimum arity. That’s because the current implementation of `case-lambda` always sets a rest argument and processed it instead of actual arguments. Which is fine! Just that I want more understanding over the structure of the lambda. Thus this pull request: to get the minimum arity of a `case-lambda` right.

Tested on this expression (try and remove some of the clauses to observe the effect:)

``` scheme
(case-lambda
 (() #f)
 ((x y z . rest) #t)
 ((x y) (+ x y))
 ((x y z) (+ (+ x y) z))
 (args #t))
```

### DIFF
--- a/lib/srfi/16.sld
+++ b/lib/srfi/16.sld
@@ -32,7 +32,7 @@
         ;; Generate the actual lambda with the right minimum arity
         ((%collect-args (shortest ...) (clauses ...))
          (lambda (shortest ... . rest)
-           (let ((len (length `((unquote shortest) ... . (unquote rest)))))
+           (let ((len (+ (length '(shortest ...)) (length rest))))
              (%case (shortest ... . rest) len 0 () clauses ...))))
         ;; Generate a lambda with a single rest argument
         ((%collect-args rest-arg (clauses ...))
--- a/lib/srfi/16.sld
+++ b/lib/srfi/16.sld
@@ -1,23 +1,61 @@
 
 (define-library (srfi 16)
   (export case-lambda)
-  (import (chibi))
+  (import (chibi)
+          (srfi 1))
   (begin
-   (define-syntax %case
-     (syntax-rules ()
-       ((%case args len n p ((params ...) . body) . rest)
-        (if (= len (length '(params ...)))
-            (apply (lambda (params ...) . body) args)
-            (%case args len 0 () . rest)))
-       ((%case args len n (p ...) ((x . y) . body) . rest)
-        (%case args len (+ n 1) (p ... x) (y . body) . rest))
-       ((%case args len n (p ...) (y . body) . rest)
-        (if (>= len n)
-            (apply (lambda (p ... . y) . body) args)
-            (%case args len 0 () . rest)))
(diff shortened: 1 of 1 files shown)


aartaka:
CC @ashinn

ashinn:
The code is much larger and harder to follow, expansion is slower, you're pulling SRFI 1 into `(scheme base)`, and the benefit seems minimal (no difference for the 0+ case and the arity alone without a proper name is not very useful).

That said, the depths to which I despise `case-lambda` cannot be expressed in words, and I'd rather not spend time on this.  Remove the SRFI 1 dependency:

```
(length* (cons* shortest ... . rest))
```

is just

```
(length `(shortest ...))
```

and I can merge this, but if it breaks I'll revert and never accept patches for `case-lambda` again :stuck_out_tongue:

aartaka:
> and the arity alone without a proper name is not very useful

Given that most `case-lambda`-s are bound to something, this should not be a problem. Signaling the right arity _in addition to_ the name is quite helpful, though!

> Remove the SRFI 1 dependency:
> (length* (cons* shortest ... . rest))
> is just
> (length `(shortest ...))

~Notice that it’s not `... . rest`, it’s `... rest` (no dot.) So the rest argument is inlined (by `cons*`) into the total arglist which is then measured by `length*`. So something like `cons*` is needed to “flatten“ the `rest` into the args.~ Nevermind, I seem to have found it:
```
`((unquote shortest) ... . (unquote rest))
```

I pushed a fixed version that no longer depends on SRFI 1. Use your judgement, I’m not particularly attached to this code and it’s fine by me if you don’t merge—it is quite repulsing indeed.

Is there maybe a better macro magic way to get the shorter of two arglists? Mine seems overly complex. Is that the way to go even?

lib/srfi/16.sld
```diff
@@ -3,21 +3,58 @@
...
+        ((%shortest-arglist (collected ...) () () cases clauses ...)
+         (%collect-args (collected ...) cases clauses ...))
+        ((%shortest-arglist (collected ...) (a as ...) (b bs ...) cases clauses ...)
+         (%shortest-arglist (collected ... a) (as ...) (bs ...) cases clauses ...))))
+    (define-syntax %collect-args
+      (syntax-rules ()
+        ;; Generate the actual lambda with the right minimum arity
+        ((%collect-args (shortest ...) (clauses ...))
+         (lambda (shortest ... . rest)
+           (let ((len (length `((unquote shortest) ... . (unquote rest)))))
```
ashinn:
```suggestion
           (let ((len (+ (length '(shortest ...)) (length rest))))
```

aartaka:
Alright, I changed my mind. This is indeed too complex of a solution for such a minor problem. `syntax-case` is prescribed in such cases instead of `syntax-rules` hacks. But I don’t want to go there. @ashinn I’m closing this.

PR #1169, opened 2026-08-14
Add -h and --help

Retropikzel:
This is small improvement for when completely new user tries to run `snow-chibi --help` or `snow-chibi -h`. Instead of error with this it makes it as same as `snow-chibi help`. If there is a better way to do this I'm all ears. :)

### DIFF
--- a/tools/snow-chibi.scm
+++ b/tools/snow-chibi.scm
@@ -238,12 +238,15 @@
       (@ ,@implementations-spec) (,command/implementations))
      (help
       "print help"
-      (,app-help-command args ...))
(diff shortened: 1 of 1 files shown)


Retropikzel:
Thank you!

ashinn:
> This is small improvement for when completely new user tries to run `snow-chibi --help` or `snow-chibi -h`. Instead of error with this it makes it as same as `snow-chibi help`. If there is a better way to do this I'm all ears. :)

Probably the app spec should allow action procedures on options, which would make this cleaner.

PR #1173, opened 2026-08-14
Memoize native-srfi-support

Retropikzel:
native-srfi-support is quite heavy for chibi and it executes quite heavy code on the target implementation too so memoizationising (what a word :smile: ) it makes things much faster.


ashinn:
> native-srfi-support is quite heavy for chibi and it executes quite heavy code on the target implementation too so memoizationising (what a word 😄 ) it makes things much faster.

"memoizing"?

Retropikzel:
Thank you!

PR #1175, opened 2026-08-16
Fix installing libraries for guile with snow-chibi

Retropikzel:
- Remove guile-srfi-fix as not needed anymore
- Use %site-ccache-dir as guile install location
- Installing .sld and .scm files is not needed by guile but left so that snow-chibi can detect if library is already installed

The code can propably be simplified and cleaned up still but now Guile finds the installed libraries.


Retropikzel:
Thank you!

PR #1176, opened 2026-08-19
Fix SRFI 231 array-decurry and array-block applying source getters more than once

### LINKED ISSUE #1105: SRFI 231: array-decurry and array-block evaluate getters multiple times for the same multi-index
I discovered a bug in my `mini-arrays.scm` implementation, wherein `array-assign!` would access each element of the source array twice, once to check with the checker and once to hand the destination setter.

I say this is a "bug" because procedures like `array-assign!`, `array-copy`, etc., might be copying or assigning an array whose elements are read sequentially from a file (like an image), so each element should be accessed only once or you'll run over the end of the image file.

So I came up with the following test:
```Scheme
(let* ((original-make-array make-array)
       (make-array
        (case-lambda
         ((specifier getter)
          (if (and (interval? specifier)
                   (procedure? getter))
              (let* ((tracker
                      (make-specialized-array specifier u1-storage-class 0))
                     (tracker_
                      (array-getter tracker))
                     (tracker!
                      (array-setter tracker))
                     (new-getter
                      (lambda args
                        (if (eqv? (apply tracker_ args) 0)
                            (begin
                              (apply tracker! 1 args)
                              (apply getter args))
                            (error "getter applied more than once: " specifier getter)))))
                (original-make-array specifier new-getter))
              ;; have original make-array handle the error
              (original-make-array specifier getter)))
         ((specifier getter setter)
          (original-make-array specifier getter setter)))))
  (let* ((A (array-copy (make-array (make-interval '#(2 2)) list) generic-storage-class #t))
         (B (make-array (make-interval '#(2 2)) list))
         (ignore (array-assign! A B))
         (B (make-array (make-interval '#(2 2)) list))
         (ignore (array-append 0 (list A B)))
         (B (make-array (make-interval '#(2 2)) list))
         (ignore (array-stack 0 (list A B)))
         (B (make-array (make-interval '#(2 2)) list))
         (ignore (array-decurry (make-array (make-interval '#(2)) (lambda (i) (list-ref (list A B) i)))))
         (B (make-array (make-interval '#(2 2)) list))
         (B* (make-array (make-interval '#(2 2)) list))
         (ignore (array-block (make-array (make-interval '#(2 2))
                                          (lambda (i j)
                                            (list-ref (list-ref (list (list A B)
                                                                      (list A B*))
                                                                i)
                                                      j)))))
         (B (make-array (make-interval '#(2 2)) list))
         (ignore (array-copy B))
         )
    #t))
```
This runs in chibi after importing `(scheme case-lambda)` and `(srfi 231)`, and it fails for both the `array-decurry` and `array-block` tests.

I looked a bit at the code and it seems that it accesses one initial element of an array to get information about the domain, etc., and then later, when it's time to copy the array elements to the result array, you access the same initial element again.

I don't have the energy right now to suggest a patch, sorry.

youdie006:
Fixes #1105.

## Problem
SRFI 231 requires that a source array's getter be applied **at most once** per multi-index -- this matters for one-shot getters, e.g. an array whose elements are read sequentially from an image file, where a second read runs past the end of the file.

`array-decurry` and `array-block` violate this:

- `array-decurry` reads the corner element once via `(apply array-ref a ...)` to determine the element domain, then reads *every* element again in `(array-for-each array-assign! curried-res a)` -- the corner is read twice.
- `array-block` reads the corner element in the initial `(assert (array? (apply get index0)))`, again in the tile-offsets loop when `i` is the lower bound, and again in the main `interval-for-each`.

Reporter @gambiteer (Bradley Lucier, the SRFI 231 author) demonstrated this with a tracker that errors "getter applied more than once"; `array-assign!` and `array-append` already pass, `array-decurry` and `array-block` fail.

## Fix
Materialize the source array once at entry with `(array-copy a generic-storage-class)`, so the original getter fires exactly once (array-copy reads each source element exactly once via `array-assign!`) and every subsequent read hits the materialized copy. The outer array holds only sub-array/tile references, so the copy is cheap, and both transforms already return eager specialized arrays so there is no change to laziness.

A more surgical single-getter variant (threading the already-read corner element through, instead of copying) is possible if you'd prefer it -- happy to switch. The materialize-once approach is the minimal, clearly-correct fix.

## Testing
Added a tracker-based regression test to `lib/srfi/231/test.sld` mirroring the reporter's approach (wrap a source array so a second read of any multi-index trips a flag), covering `array-assign!` and `array-append` as passing controls plus `array-decurry` and `array-block` as the regressions.

- Before the fix: the two new cases FAIL (getter applied more than once).
- After the fix: all pass. Full SRFI 231 suite: 579/579 tests pass, no regression.

Credit to @gambiteer for the report and the tracker test.

---
This change was prepared with AI assistance and reviewed by me before submission.

### DIFF
--- a/lib/srfi/231/test.sld
+++ b/lib/srfi/231/test.sld
@@ -3359,6 +3359,70 @@
               (make-array (make-interval '#(0)) list))))
         )
 
+      ;; https://github.com/ashinn/chibi-scheme/issues/1105
+      ;; SRFI 231 requires a source array's getter be applied at most once
+      ;; per multi-index (matters for one-shot getters such as reading the
+      ;; elements of an image sequentially from a file).  array-decurry and
+      ;; array-block used to read the corner element more than once.  This
+      ;; mirrors the tracker test from the reporter, @gambiteer (Bradley
+      ;; Lucier, the SRFI 231 author): wrap a source array so a second read
+      ;; of any multi-index trips a flag, then assert the flag stays clear.
+      (test-group "getter applied at most once (issue #1105)"
+        (let ()
+          ;; Wrap arr over the same domain; set (vector-ref flag 0) to #t
+          ;; if any multi-index is read more than once.
+          (define (track arr flag)
+            (let* ((domain (array-domain arr))
+                   (seen (make-specialized-array domain u1-storage-class 0))
+                   (seen_ (array-getter seen))
+                   (seen! (array-setter seen))
+                   (get (array-getter arr)))
+              (make-array
+               domain
+               (lambda multi-index
+                 (if (not (eqv? (apply seen_ multi-index) 0))
+                     (vector-set! flag 0 #t))
+                 (apply seen! 1 multi-index)
+                 (apply get multi-index)))))
+          ;; array-assign! and array-append already satisfy the requirement.
+          (test-assert "array-assign! reads each source element at most once"
+            (let* ((flag (vector #f))
+                   (dest (array-copy (make-array (make-interval '#(2 2)) list)
+                                     generic-storage-class #t))
+                   (src (track (make-array (make-interval '#(2 2)) list) flag)))
+              (array-assign! dest src)
+              (not (vector-ref flag 0))))
+          (test-assert "array-append reads each source element at most once"
+            (let* ((flag (vector #f))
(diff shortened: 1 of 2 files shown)


ashinn:
Thanks!  I've copied your tests as-is, and fixed this with an optimization for the specialized-array input case.

PR #1182, opened 2026-08-26
Add guile-srfi-fix back into snow-chibi

Retropikzel:
It seems that I was wrong about Guile not needing the srfi-fixes. Apologies. (include ...) not working like in other (and it is all other) R7RS implementations is fixed now https://codeberg.org/guile/guile/issues/139#issuecomment-21700217 so I'm testing on latest guile head on which these changes now work on.


Retropikzel:
Thank you!

PR #1183, opened 2026-08-27
fix reading of \x80; in string

mcgoron-peter:
Writing "\x80;" left one with a corrupted string because it wasn't decoded properly. This caused strange issues like `(string-length "\x80;") => 1` and `(string->list "\x80;") => 0`.


ashinn:
Wow, good catch.

PR #1189, opened 2026-08-30
Another attempt to generate smarter arity-inspectable case-lambdas

aartaka:
Yes, I’m at it again—trying to make `case-lambda`-generated lambdas to have more inspectable and truthful arities. This time, the implementation is short (17 lines -> 30 lines,) tested, documented, and more straightforward in general. Expansion time shouln’t be too long either, especially compared to the previous attempt (#1167), where a whole set of recursive macros handling arglists was used.

@ashinn how does that look?

### DIFF
--- a/lib/srfi/16.sld
+++ b/lib/srfi/16.sld
@@ -26,6 +26,8 @@
          (%case-lambda (shortest ... (arg ...)) (args ...) clauses ...))
         ;; When ran out, generate a lambda with shortest args followed
         ;; by rest arg
+        ;; This (arg . _) is here because we need to “flatten” the
+        ;; nested list
         ((%case-lambda ((arg . _) ...) non-parsed-args clauses ...)
          (lambda (arg ... . rest)
            (let ((len (length `((unquote arg) ... . (unquote rest)))))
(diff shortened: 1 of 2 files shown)


lib/srfi/16.sld
```diff
@@ -3,21 +3,41 @@
...
+    (define-syntax %case-lambda
+      (syntax-rules ()
+        ;; Walk args until common ones run out
+        ((%case-lambda (shortest ...) ((arg . args) ...) clauses ...)
+         (%case-lambda (shortest ... (arg ...)) (args ...) clauses ...))
+        ;; When ran out, generate a lambda with shortest args followed
+        ;; by rest arg
+        ((%case-lambda ((arg . _) ...) non-parsed-args clauses ...)
+         (lambda (arg ... . rest)
+           (let ((len (length `((unquote arg) ... . (unquote rest)))))
```
ashinn: Currently Chibi doesn't constant fold `length` on a constant list but I'm likely to implement that. Even without the folding, there's no reason to construct a list at runtime here:

```suggestion
           (let ((len (+ (length '(arg ...)) (length rest))))
```

aartaka:
@ashinn I believe I fixed the length calculation

Issue #1185, opened 2026-08-29
Ensuring correct .snow-repo.scm hashes


Retropikzel:
I've started to try and work mainly with git based repositories with snow-chibi. And I run into an usability issue of short. As an experienced idiot myself, I have already managed to foorgun myself with this multiple times without noticing. 😄 

Say you develop library with files foo/bar.sld and foor/bar.scm. You make changes to them but do not commit them to git. You then package the new version of your library with `snow-chibi package --version=1.0.1 foo/bar.sld` and run `snow-chibi git-index foo-bar-1.0.1.tgz` which updates the .snow-repo.scm. Except that the hash in .snow-repo.scm points to current git hash, which do not yet have the changes to foo/bar.sld and foo/bar.scm in it. So the .snow-repo.scm is pointing to the older wrong hash.

I think `snow-chibi git-index` should run `git status -s` and check that if it reports any changed files. If it does then issue warning about possibly wrong hash. Later more advanced versions could propably issue this warning only if files inside the package are in the output of `git status -s`.

For those who do not want to make the .snow-repo.scm updating it's own commit I propose adding `--tag` option to git-index command. If it is given then the hash is not read from git log and the tag is used instead. User can then do:
```
snow-chibi package --version=1.0.1 foo/bar.sld
snow-chibi git-index --tag=v1.0.1 foo-bar-1.0.1.tgz
git add .
git commit -m "Version 1.0.1"
git push
git tag v1.0.1
git push --tags
```

@ashinn What do you think?

ashinn:
sgtm

PR #1191, opened 2026-09-02
Snow chibi git index warning

Retropikzel:
When git indexing add warning if `git status -s` reports anything. It's purpose is to prevent user from indexing files into .snow-repo.scm that are not commited into git and/or wrong git commit hash ending up in .snow-repo.scm.


Retropikzel:
Thank you!

Retropikzel:
https://github.com/ashinn/chibi-scheme/issues/1185

PR #1194, opened 2026-09-05
Use SRFI-145 assume in SRFI-253 code

aartaka:
Thanks to using `assume` instead of `assert`, the checks can now be stripped off in “production” mode. Important note: tests for SRFI-253 and its dependant 273 now require a `debug` feature on, but I don’t know how to enable that for an individual test :confused:

### DIFF
--- a/lib/srfi/253.sld
+++ b/lib/srfi/253.sld
@@ -2,9 +2,9 @@
   (import (scheme base)
           (scheme case-lambda)
           (scheme list)
-          (chibi assert)
           (chibi ast)
           (chibi generic))
+  (import (srfi 145))
   (export check-impl?
           check-arg values-checked
           check-case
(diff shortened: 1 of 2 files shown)


aartaka:
CC @ashinn

ashinn:
I think it's not appropriate for Chibi to elide checks here by default, because there is no static analysis being done.

aartaka:
@ashinn hmmm, but what about performance?

ashinn:
I mean, it's your SRFI so it's up to you. I don't know what the users would typically expect.

If you want this change you have to make the tests pass first, though, even if just disabling the affected ones.

aartaka:
@ashinn is it possible to add `debug` feature for just one test then?

ashinn:
> @ashinn is it possible to add `debug` feature for just one test then?

No, the tests are all compiled together before they are run.

aartaka:
> > @ashinn is it possible to add `debug` feature for just one test then?
> 
> No, the tests are all compiled together before they are run.

Damn. Okay, closing this PR then! It was questionable in the first place.

PR #1199, opened 2026-09-08
Handle HTTP request parse errors and close rejected uploads cleanly

jordanlhenderson:
Requests such as `POST /?a&b=c` and `POST /?&` can fail during request construction, and malformed query escapes such as `/?x=%GG` return HTTP 500 before a servlet runs. A servlet that rejects a POST without reading its body can also lose its response to a TCP reset.

This change:

- Compares the position of `=` with the current query field's end, so bare and empty fields parse consistently wherever they occur.
- Returns a complete HTTP 400 response when request construction fails. Unhandled servlet exceptions still return 500.
- Half-closes server output after flushing the response, then drains unread socket input until EOF or a one-second deadline before closing the socket. This follows the staged close described in [RFC 9112, section 9.6](https://www.rfc-editor.org/rfc/rfc9112.html#section-9.6), uses a fixed-size buffer, and does not depend on client-supplied body framing.
- Preserves a waiting thread's existing deadline when the scheduler requeues it. Replacing that deadline with an indefinite wait prevented the socket drain from timing out when a peer left its write side open.

Adds URI boundary regressions and a loopback `make test-net` suite, included in `make test-all`. The network tests cover malformed queries, servlet dispatch, intentional servlet failures, unread 8 KiB and 128 KiB bodies, complete responses with orderly EOF, and cleanup when the peer remains open. They use Chibi itself, with no external HTTP client dependency.

Validation on Linux:

- `make -j2` and `make test-all` pass, including 64 URI assertions and 28 network assertions.
- The idle-peer regression failed without the scheduler change and passes with it.


ashinn:
Thanks, haven't forgotten this, just been busy!

PR #1200, opened 2026-09-10
Allow specifying tag when git-indexing

Retropikzel:
Allow specifying git tag when indexing. If --tag is given then git repo is not checked for uncommitted changes and no hash is added to the .snow-repo.scm package definition. Git-indexing still has to happen in git repository as url is read from there. Propably good idea to allow taking that as argument at one point too, that would allow easy decoupling of the source git repository and snow index git repository. It of course can now be done by hand.


Retropikzel:
https://github.com/ashinn/chibi-scheme/issues/1185

Issue #1196, opened 2026-09-07
SRFI 231: interval-cursor-next! doesn't play well with continuations


gambiteer:
This test started with the example in https://github.com/ashinn/chibi-scheme/issues/904#issuecomment-5561546143 , was modified by the Kaappi Scheme author/AI, and then modified back by me.
```
(test-group
 "continuation tests"
 
 (define (re-entry-results collect)
   (let* ((cont1 #f) (cont2 #f) (i 5) (first-run? #t)
          (f (lambda (x)
               (call-with-current-continuation
                (lambda (c)
                  (if first-run?
                      (case x ((1) (set! cont1 c)) ((3) (set! cont2 c)) (else #f)))
                  x))))
          (results '()))
     (let ((r (collect f)))
       (set! first-run? #f)
       (set! results (cons r results)))
     (case i
       ((5) (set! i (- i 1)) (cont1 10))
       ((4) (set! i (- i 1)) (cont1 20))
       ((3) (set! i (- i 1)) (cont2 10))
       ((2) (set! i (- i 1)) (cont2 20))
       (else #t))
     results))

 (define re-entry-expected '((0 1 2 20) (0 1 2 10) (0 20 2 3) (0 10 2 3) (0 1 2 3)))

 (define (over-f f) (make-array (make-interval '#(4)) f))

 (test "array-fold-left threads its accumulator functionally"
       re-entry-expected
       (re-entry-results (lambda (f) (reverse (array-fold-left (lambda (acc x) (cons x acc)) '() (over-f f))))))
 (test "array-fold-right threads its accumulator functionally"
       re-entry-expected
       (re-entry-results (lambda (f) (array-fold-right cons '() (over-f f)))))
 (test "array-reduce threads its accumulator functionally"
       '(23 13 25 15 6)
       (re-entry-results (lambda (f) (array-reduce + (over-f f)))))
 (test "array->list threads its accumulator functionally"
       re-entry-expected
       (re-entry-results (lambda (f) (array->list (over-f f)))))
 (test "array->vector threads its accumulator functionally"
       (map list->vector re-entry-expected)
       (re-entry-results (lambda (f) (array->vector (over-f f)))))
 (test "array-every threads its accumulator functionally"
       '((last 20) (last 10) 3 3 3)
       (re-entry-results
        (lambda (f) (array-every (lambda (x) (if (>= x 10) (list 'last x) x)) (over-f f)))))
 )
```
[...]

PR #1201, opened 2026-09-10
SRFI 231: Use a functional approach to step through interval multi-ind…

gambiteer:
…ices

Add a routine make-reverse-index-cursor that takes an nterval argument and returns a procedure that,
given a reversed multi-index in that interval, returns the next reversed multi-index in that interval, or #f if there are no more multi-indices in that interval

Use the new procedure to step through the multi-indices of an interval in interval-fold-left when the dimension is > 2, and interval-fold-right for all dimensions.

Also use it in array-assign/reshape!, which is called only by specialized-array-reshape when that routine needs to copy the argument array.

### DIFF
--- a/lib/srfi/231/base.scm
+++ b/lib/srfi/231/base.scm
@@ -124,42 +124,25 @@
             (make-interval (vector-copy (interval-lb iv) (- n rd))
                            (vector-copy (interval-ub iv) (- n rd))))))
 
-(define (rev-index-next! rev-index rev-lowers rev-uppers)
-  (cond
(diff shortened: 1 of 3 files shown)


gambiteer:
Addresses https://github.com/ashinn/chibi-scheme/issues/1196

gambiteer:
I plan a couple more small fixes, so don't apply yet.

gambiteer:
These are all the changes I intend to make.

ashinn:
I want to play with this some more but am tentatively merging for now.

